PR #1, opened 2016-01-20
Bring in the latest 

latrelstack:
Rpm creation capability  for px-fuse


latrelstack:
Rpm packaging creation capability and scripts.

PR #4, opened 2017-05-01
Few trivial fixes for pxd.c

sashalevin:


pxd.c
```diff
@@ -1132,7 +1132,7 @@ int pxd_init(void)
 out_misc:
 	misc_deregister(&pxd_miscdev);
 out_fuse:
-	for (j = 0; j < i; ++j) {
+	for (j = 0; j <= i; ++j) {
```
prabirpaul: Context at i need to be destroyed only if we are here from line: 1101. Else context at i is either uninitialized or beyond limit. You could just add a pxd_context_destroy() at line: 1101.

sashalevin: I could, but isn't this simpler? This also keeps the failure handling logic in one place (the error labels) rather than sprinkling them all over the function.

prabirpaul: Yes it is. But in the for loop above (line: 1090), if call to pxd_context_init() fails, context at i will be uninitialized.

sashalevin: Maybe instead the init function should be refactored so it wouldn't be allowed to fail?

sashalevin:
Re-pushed (I hope it works right on github).

pxd.c
```diff
@@ -1035,23 +1035,14 @@ static void pxd_timeout(unsigned long args)
...
-
-	if (ctx->id < pxd_num_contexts_exported) {
-		err = fuse_conn_init(&ctx->fc);
-		if (err)
-			return err;
-	} else {
-		ctx->fops.unlocked_ioctl = pxd_control_ioctl;
-	}
 	ctx->fc.release = pxd_fuse_conn_release;
 	ctx->fc.allow_disconnected = 1;
```
prabirpaul: These two need to move to pxd_context_setup(). fuse_conn_init() is going to zero out ctx->fc.

pxd.c
prabirpaul: I think problem still exists. If pxd_context_setup() fails at i, we cant call destroy for i.

PR #6, opened 2017-05-02
Close up pxd_dev leaks

sashalevin:


pxd.c
```diff
@@ -605,6 +605,7 @@ ssize_t pxd_add(struct fuse_conn *fc, struct pxd_add_out *add)
 out_id:
 	ida_simple_remove(&pxd_minor_ida, new_minor);
 out_module:
+	kfree(pxd_dev);
```
prabirpaul: pxd_dev could be null here (line: 558). Add this free to pxd_free_disk(), that would cover all our bases.

sashalevin: kfree(NULL) is defined to not do anything.

Moving that kfree() into pxd_free_disk() will leave sites it can leak from, for example this very same pxd_add():

pxd_dev = kzalloc(sizeof(*pxd_dev), GFP_KERNEL);
[...]
new_minor = ida_simple_get(&pxd_minor_ida, 1, 1 << MINORBITS, GFP_KERNEL);
[ this fails]
goto out_module;
[ leak ]

prabirpaul: ok, that works.

PR #9, opened 2018-05-30
- Version upgrade

prabirpaul:
- Safe free during pxd add


pxd.c
```diff
@@ -654,7 +654,8 @@ ssize_t pxd_add(struct fuse_conn *fc, struct pxd_add_out *add)
 out_id:
 	ida_simple_remove(&pxd_minor_ida, new_minor);
 out_module:
-	kfree(pxd_dev);
+	if (pxd_dev)
```
maxkozlovsky: the check for null is redundant, kfree checks

PR #16, opened 2018-11-27
**** extend pxd to use inkernel path for IO ******

sulakshm:
Make pxd to take inkernel IO path instead of fuse.
Please note, the changes has been compiled upto latest kernels upto 4.4, but only tested with 3.10.x kernels ("CentOS Linux 7 (Core)")

Dependent changes:
https://github.com/portworx/porx/pull/2698


sulakshm:
Not ready yet.

PR #18, opened 2019-01-03
Pxd merging

jjobi:
couple of changes for feedback... may need more verification...

1.  Merge sequential write requests in pxd driver.
2.  reset active_background counter immediately after signaling so that PX won't be signaled again before accumulating enough requests.

### DIFF
--- a/dev.c
+++ b/dev.c
@@ -211,9 +211,9 @@ static void queue_request(struct fuse_conn *fc, struct fuse_req *req)
 	req->state = FUSE_REQ_PENDING;
 }
 
-static void fuse_conn_wakeup(struct fuse_conn *fc)
+static void fuse_conn_wakeup(struct fuse_conn *fc, bool force)
 {
-    if (!fc->signaled) {
+    if (!fc->signaled || force) {
         fc->signaled = true;
 	    wake_up(&fc->waitq);
 	    kill_fasync(&fc->fasync, SIGIO, POLL_IN);
@@ -235,7 +235,7 @@ void fuse_request_send_oob(struct fuse_conn *fc, struct fuse_req *req)
 	req->state = FUSE_REQ_PENDING;
 	spin_unlock(&fc->lock);
 
-	fuse_conn_wakeup(fc);
+	fuse_conn_wakeup(fc, false);
 }
 
 /*
@@ -299,7 +299,6 @@ static inline bool fuse_request_queue(struct fuse_conn *fc,
         ((req->in.h.opcode == PXD_WRITE) &&
          ((req->misc.pxd_rdwr_in.size == 0) ||
           (req->misc.pxd_rdwr_in.flags & PXD_FLAGS_FLUSH)))) {
-        fc->active_background = 0;
         return true;
     }
     return false;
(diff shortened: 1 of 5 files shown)


pxd.c
```diff
@@ -1118,6 +1125,25 @@ static void pxd_fuse_conn_release(struct fuse_conn *conn)
 {
 }
 
+static void pxd_wakeup(struct pxd_context *ctx)
+{
```
sulakshm: It seems to me this newly introduced function pxd_wakeup from the timer, is running without any synchronization with the main IO path... so how would all the state variables used below be consistent?

dev.c
```diff
@@ -262,17 +265,48 @@ __releases(fc->lock)
...
 {
+    struct fuse_req *prev;
+    bool wakeup = false;
+
+    /* Check if the request could be merged with another request */
+    if (req->in.h.opcode == PXD_WRITE && !list_empty(&fc->pending)) {
+        prev = list_last_entry(&fc->pending, struct fuse_req, list);
+        if ((prev->in.h.opcode == PXD_WRITE) &&
+            ((prev->misc.pxd_rdwr_in.offset + prev->misc.pxd_rdwr_in.size) ==
+             req->misc.pxd_rdwr_in.offset) &&
```
sulakshm: the above seems to do only forward sequential merge? how about backward merge? If A1, A2 are continuous blocks, then A1->A2 is handled, A2->A1 seems not handled!

dev.c
```diff
@@ -281,12 +315,24 @@ static void fuse_request_send_nowait_locked(struct fuse_conn *fc,
 		set_bdi_congested(&fc->bdi, BLK_RW_ASYNC);
 	}
 	fc->active_background++;
+    if (fc->active_background >= fc->accumulate) {
+        fc->active_background = 0;
+        wakeup = true;
+    }
 	req->in.h.unique = fuse_get_unique(fc);
 	queue_request(fc, req);
```
sulakshm: If the requests miss the simple sequential merge above, then they are added to the tail... This impacts the merging logic above, if the requests are ordered in ascending order of block addresses, then it is possible to do more merges when out of order requests gets submitted/processed.

dev.c
```diff
@@ -262,17 +265,48 @@ __releases(fc->lock)
...
+static bool fuse_request_send_nowait_locked(struct fuse_conn *fc,
 					    struct fuse_req *req)
 {
+    struct fuse_req *prev;
+    bool wakeup = false;
+
+    /* Check if the request could be merged with another request */
+    if (req->in.h.opcode == PXD_WRITE && !list_empty(&fc->pending)) {
+        prev = list_last_entry(&fc->pending, struct fuse_req, list);
+        if ((prev->in.h.opcode == PXD_WRITE) &&
```
sulakshm: why only restrict to writes? sequential reads do help too with readaheads happening.

dev.c
jjobi: Yes, this could be enhanced a lot more... This specific change was just for demonstrating the benefit of merging requests in pxd driver.

dev.c
jjobi: We could maintain a sorted list in pxd driver, but our current code was not even doing good job when the kernel was sending writes in order.

dev.c
jjobi: I am testing changes for merging reads and ran into some issues...  This is work in progress...

pxd.c
jjobi: I think we don't have to add much synchronization for the timer since the timer is fired so frequently.  As you might have noticed, I am working on top of Vinod's changes.

jjobi:
Moved the changes to another branch.  A separate PR will be created.

PR #19, opened 2019-01-09
Convert writes of zeros to discard request

jjobi:
PXD driver can convert writes of zero blocks as discard requests and avoid unnecessary transfer of zeroes from kernel to user space.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -303,7 +303,37 @@ static void pxd_read_request(struct fuse_req *req, uint32_t size, uint64_t off,
 static void pxd_write_request(struct fuse_req *req, uint32_t size, uint64_t off,
 			uint32_t minor, uint32_t flags, bool qfn)
 {
+	struct req_iterator breq_iter;
+#ifdef HAVE_BVEC_ITER
+	struct bio_vec bvec;
+#else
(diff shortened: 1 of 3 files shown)


pxd.c
```diff
@@ -303,7 +303,37 @@ static void pxd_read_request(struct fuse_req *req, uint32_t size, uint64_t off,
...
+
+	/* Check if this is a write of zero blocks and if so, convert that to a
+	 * discard request.
+	 */
+	if (size && qfn && !(flags & (REQ_FLUSH | REQ_FUA))) {
+		rq_for_each_segment(bvec, req->rq, breq_iter) {
+			kaddr = kmap_atomic(BVEC(bvec).bv_page);
+			p = kaddr + BVEC(bvec).bv_offset;
+			len = BVEC(bvec).bv_len;
+			for (i = 0; i < len; i++) {
```
prabirpaul: Looks pretty expensive. Aren't there any cpu intrinsics you can use for this, or compare word at a time?

kschoche: You can use memcmp (not sure if optimized version is available in all kernel versions) or go 8 bytes at a time comparing to 0ULL and just handle the alignment.

PR #21, opened 2019-01-12
Try to pick up as many requests in every iteration

jjobi:
Try to pick as many requests each time requests are copied from kernel to user space.  This seems to improve our throughput.


dev.c
```diff
@@ -468,6 +468,16 @@ static ssize_t fuse_dev_do_read(struct fuse_conn *fc, struct file *file,
 		}
 	}
 
+	/* Check if more requests could be picked up */
+	if (remain && request_pending(fc)) {
```
prabirpaul: Instead  of 2 checks for request_pending() here you can probably just jump to L397 and make the necessary changes under the existing request_pending() check.

PR #26, opened 2019-03-14
ln/pxdkernel-1: use direct bio instead of request queue for blkdev.

sulakshm:
This is a series of changes to bring in pxdkernel changes into master.

- use direct bio instead of request queue for blkdev. 

a) The real optimization happens with the actual target device (sorting, queueing etc..). For px virtual device, it is best to pick and punt at the earliest.
b) Using bio will also allow to switch (zero copy) if the target device is a block device as well.


sulakshm:

PR #32, opened 2019-03-19
ln/pxdkernel-5: provide a path to configure backing device/file dynamically

sulakshm:
This changeset part of the kernel px driver, that enables fast path operation to kick in, once a backing file/device is configured through the extended interface.


sulakshm:

PR #34, opened 2019-05-15
misc blk-mq changes and cleanup

prabirpaul:


pxd.c
```diff
@@ -632,21 +613,20 @@ static int pxd_init_disk(struct pxd_device *pxd_dev, struct pxd_add_out *add)
...
 		blk_queue_make_request(q, pxd_make_request);
 	} else {
 #ifdef __PX_BLKMQ__
-	  int err;
-
 	  memset(&pxd_dev->tag_set, 0, sizeof(pxd_dev->tag_set));
 	  pxd_dev->tag_set.ops = &pxd_mq_ops;
-
-
 	  pxd_dev->tag_set.queue_depth = PXD_MAX_QDEPTH;
```
jrivera-px: tag_set->'queue_depth' is set here to PXD_MAX_QDEPTH and also set at line 629 to '128'?    We want PXD_MAX_QDEPTH or '128'?

PR #35, opened 2019-05-15
Blkmq support

prabirpaul:


prabirpaul:
blk-mq on Kernel 5.0 might need some more changes. It will be handled later.

PR #36, opened 2019-05-17
PWX-8766: Use __PX_BLKMQ__ define in pxd_update_stats.

jrivera-px:
pxd_update_stats uses, 'part_stat_inc' & 'part_stat_add'.  Those functions changes in EL8.    Need to use this define to get around them for now.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -194,12 +194,19 @@ static const struct block_device_operations pxd_bd_ops = {
 
 static void pxd_update_stats(struct fuse_req *req, int rw, unsigned int count)
 {
-	struct pxd_device *pxd_dev = req->queue->queuedata;
+        struct pxd_device *pxd_dev = req->queue->queuedata;
 
-	int cpu = part_stat_lock();
-	part_stat_inc(cpu, &pxd_dev->disk->part0, ios[rw]);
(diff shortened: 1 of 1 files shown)


pxd.c
```diff
@@ -194,12 +194,19 @@ static const struct block_device_operations pxd_bd_ops = {
...
 static void pxd_update_stats(struct fuse_req *req, int rw, unsigned int count)
 {
-	struct pxd_device *pxd_dev = req->queue->queuedata;
+        struct pxd_device *pxd_dev = req->queue->queuedata;
 
-	int cpu = part_stat_lock();
-	part_stat_inc(cpu, &pxd_dev->disk->part0, ios[rw]);
-	part_stat_add(cpu, &pxd_dev->disk->part0, sectors[rw], count);
-	part_stat_unlock();
+#ifdef __PX_BLKMQ__
```
jrivera-px: Actually we can change this #ifdef __PX_BLKMQ__ to #ifdef part_stat_get_cpu.   The define 'part_stat_get_cpu' seems to have been added with the other part_stat changes.    Should we try that one?

prabirpaul: Let's merge this in and then try anything new.

PR #38, opened 2019-06-28
fix for small sync io perf issue

sulakshm:
This PR addresses perf issue for small sequential IO in sync mode.

The test profile is defined as 128MB sequential write with fdatasync set.
This test profile, opens the target file in O_RDWR|O_SYNC|O_DIRECT|O_NOATIME
This open mode implies that fdatasync is actually a nop.
Also, the queue depth is just 1. This implies that progress gets made one 
request at a time.

The behavior though is as below.
  write: io=8227.5MB, bw=140413KB/s, iops=35103, runt= 60001msec

With px virtual device, there is 2 hop in the IO path, one from user app (fio) to the
virtual px device (/dev/pxd) and then within the px driver to the actual target file 
(/var/.px/0/uuid/pxdev). So atleast 50% of direct ext4 file io iops/throughput is minimal
expectation.

Inside, in the px driver, we open the target file in O_RDWR|O_DIRECT|O_LARGEFILE.
So could file mode have any impact?

Secondly, writes schedule further processing on threads in the same cpu.
Could this be a bottleneck? Yes. If the kernel threads are unbound, then they 
are free to be scheduled on any idle cpu, which pushes throughput higher.
This one change is significant to cause throughput to reach 60% of direct ext4.

direct ext4 - 34K iops
  write: io=8073.4MB, bw=137783KB/s, iops=34445, runt= 60001msec

fast path - 
  write: io=1488.4MB, bw=25401KB/s, iops=6350, runt= 60001msec
  write: io=1602.2MB, bw=27343KB/s, iops=6835, runt= 60001msec

slow path - 
  write: io=393528KB, bw=6558.8KB/s, iops=1639, runt= 60001msec

fast path, with thread unbound
  write: io=4975.4MB, bw=84912KB/s, iops=21227, runt= 60001msec

fastpath: sync + noatime + thread unbound - 21k
  write: io=4965.3MB, bw=84739KB/s, iops=21184, runt= 60001msec
  write: io=5042.8MB, bw=86061KB/s, iops=21515, runt= 60001msec

Below are the changes specific to this test case:
1/ update file open mode of target file.

2/ if target file is opened in O_SYNC with px driver, then any sync call 
on that file shall become nop. this can become another arg part of the pxd_add_device ioctl.

3/ there is one kthread bound on cpu, to service writes on the same cpu on which it is received.
This introduces scheduling delay on the cpu. If the kthreads are left unbounded, then 
it increases chances of thread being scheduled on other idle cpu, and improves throughput.
The correct fix on this should be to create a pool of idle threads per cpu and rotate among
them, instead of just one dedicated thread. 

4/ while we were on it, i added few more counters to dump io statistics in more detail.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -554,6 +554,7 @@ static int pxd_init_disk(struct pxd_device *pxd_dev, struct pxd_add_out *add)
 	blk_queue_logical_block_size(q, PXD_LBS);
 	blk_queue_physical_block_size(q, PXD_LBS);
 	//blk_queue_flag_set(QUEUE_FLAG_NOMERGES, q);
+	//queue_flag_set_unlocked(QUEUE_FLAG_NOMERGES, q);
 
 	set_capacity(disk, add->size / SECTOR_SIZE);
 
@@ -789,6 +790,10 @@ ssize_t pxd_update_path(struct fuse_conn *fc, struct pxd_update_path_out *update
 	struct pxd_device *pxd_dev;
 	int i;
 	struct file* f;
+	int mode = O_DIRECT | O_LARGEFILE | O_RDWR | O_NOATIME;
+
+	printk(KERN_INFO "%s for path %s\n",
+		__func__, update_path->devpath[0]);
 
 	spin_lock(&ctx->lock);
 	list_for_each_entry(pxd_dev, &ctx->list, node) {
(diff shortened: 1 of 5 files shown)


sulakshm:

PR #51, opened 2019-11-08
remove some unused code and write only fields

maxkozlovsky:


fuse_i.h
```diff
@@ -391,25 +382,9 @@ struct fuse_conn {
...
 	struct rb_root polled_files;
 
-	/** Maximum number of outstanding background requests */
-	unsigned max_background;
-
-	/** Number of background requests at which congestion starts */
-	unsigned congestion_threshold;
-
-	/** Number of requests currently in the background */
-	unsigned num_background;
```
prabirpaul: No need to remove this counter, helps with debugging.

maxkozlovsky: I don't see how and what particular problem this could possibly help to debug. There is nowhere for requests to go.

PR #52, opened 2019-11-11
special case pxd init handling

maxkozlovsky:
remove page descriptor handling from requests
pxd init was the only request type that was using the fields

### DIFF
--- a/dev.c
+++ b/dev.c
@@ -25,17 +25,6 @@
 #include <linux/blkdev.h>
 #include "pxd_compat.h"
 
-#if LINUX_VERSION_CODE < KERNEL_VERSION(3,18,0)
-#include "iov_iter.h"
-
-#define iov_iter_advance __iov_iter_advance
-#define iov_iter __iov_iter
-#define iov_iter_init __iov_iter_init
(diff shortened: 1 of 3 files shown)


maxkozlovsky:
init is the only request using page descriptors, which makes every data requests do unnecessary work on maintaining unused fields, increases the size of data structure, etc. etc.

What does "line up" mean. There can be only one open/close, the next close lines up with previous open. You don't need any additional counters for this.

prabirpaul:
How does that justify adding exception cases in fuse read/write path. Put it in the union if all u want is to save 2 ptrs.

Not when you have bugs. I have used the unique on multiple customer cases to check if multiple init requests are running.

maxkozlovsky:
You are asking wrong question. The right question is how do you justify adding extra code to data path when all you need is somehow doing init. If the page fields were not present and the person implementing init would try to add them, it obviously would not be accepted. This justifies removing the fields and doing something else for init. Init will not be in the read/write path eventually, it'll be an ioctl when I have enough changes to have driver version bumped.

Fine I'll add your counters back.

PR #57, opened 2019-11-21
use ioctl instead of request for init call

maxkozlovsky:
remove init handling from data path
bump module version 

https://github.com/portworx/porx/pull/4856 for corresponding user space changes


jvinod:
When this moves to porx, will we retain init call in userspace for backward compatibility or will upgrade require reboot

maxkozlovsky:
It's backward compatible, see https://github.com/portworx/porx/pull/4856

PR #58, opened 2019-11-22
Add define to denote el8 kernels so 5.x blq mq changes can be used fo…

jrivera-px:
…r el8.  This is just the start.  Will need to replace '__PX_BLKMQ__' with '#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,0,0) || __EL8__' in other places of the code.

Signed-off-by: Jose Rivera <jose@portworx.com>

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -50,7 +50,7 @@ export OUTPATH
 
 # EL8 Specific kernel checks
 ifeq ($(shell echo "$(CHK_KVER)" | grep -Eq '.*\.el8.*\.x86_64'; echo $$?),0)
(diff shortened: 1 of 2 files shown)


maxkozlovsky:
This does not work with newer compilers:

/home/max/jenkins/workspace/Porx-UT/go/src/github.com/portworx/porx/src/gdfs/px_fuse/pxd.c:191:52: error: "__EL8__" is not defined, evaluates to 0 [-Werror=undef]
20:28:37  #if LINUX_VERSION_CODE >= KERNEL_VERSION(5,0,0) || __EL8__

You need a value for the define or the check should be || defined(__EL8__)

PR #62, opened 2019-12-04
use array to lookup request by id

maxkozlovsky:
this does not require locks while hash does
per cpu id allocators

### DIFF
--- a/dev.c
+++ b/dev.c
@@ -36,7 +36,7 @@
 /** Maximum number of outstanding background requests */
 #define FUSE_DEFAULT_MAX_BACKGROUND (PXD_MAX_QDEPTH * PXD_MAX_DEVICES)
 
-#define FUSE_HASH_SIZE FUSE_DEFAULT_MAX_BACKGROUND
+#define FUSE_MAX_REQUEST_IDS (2 * FUSE_DEFAULT_MAX_BACKGROUND)
 
 static struct kmem_cache *fuse_req_cachep;
 
@@ -53,7 +53,6 @@ void fuse_request_init(struct fuse_req *req)
 {
 	memset(req, 0, sizeof(*req));
 	INIT_LIST_HEAD(&req->list);
-	INIT_HLIST_NODE(&req->hash_entry);
 }
 
 static struct fuse_req *__fuse_request_alloc(gfp_t flags)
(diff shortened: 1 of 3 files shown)


dev.c
```diff
@@ -135,20 +134,70 @@ static unsigned len_args(unsigned numargs, struct fuse_arg *args)
...
+		spin_unlock(&fc->lock);
+
+		my_ids->num_free_ids = num_alloc;
+	}
+
+	uid = my_ids->free_ids[--my_ids->num_free_ids];
+
+	put_cpu();
+
+	uid += FUSE_MAX_REQUEST_IDS;
```
prabirpaul: I don't get this. We just retrieved a new uid from a stack of existing uids initialized at L891. Why the + FUSE_MAX_REQUEST_IDS. And also where's the corresponding -FUSE_MAX_REQUEST_IDS when putting back?

dev.c
```diff
@@ -135,20 +134,70 @@ static unsigned len_args(unsigned numargs, struct fuse_arg *args)
...
+
 	/* zero is special */
-	if (unlikely(fc->reqctr == 0))
-		fc->reqctr = 1;
+	if (uid == 0)
```
prabirpaul: This can can never be true because of L161 above.

dev.c
```diff
@@ -135,20 +134,70 @@ static unsigned len_args(unsigned numargs, struct fuse_arg *args)
...
+	u64 uid;
+	int num_alloc;
+
+	int cpu = get_cpu();
+
+	my_ids = per_cpu_ptr(fc->per_cpu_ids, cpu);
+
+	if (unlikely(my_ids->num_free_ids == 0)) {
+		spin_lock(&fc->lock);
+		BUG_ON(fc->num_free_ids == 0);
```
prabirpaul: What guarantees fc->num_free_ids != 0? Why wouldn't u run out of ids if u have a large number of cpus and a subset of them have consumed most of the free_ids. And fuse_get_unique() get called on a cpu which doesn't have free_ids stack yet?

dev.c
```diff
@@ -36,7 +36,7 @@
 /** Maximum number of outstanding background requests */
 #define FUSE_DEFAULT_MAX_BACKGROUND (PXD_MAX_QDEPTH * PXD_MAX_DEVICES)
 
-#define FUSE_HASH_SIZE FUSE_DEFAULT_MAX_BACKGROUND
+#define FUSE_MAX_REQUEST_IDS (2 * FUSE_DEFAULT_MAX_BACKGROUND)
```
prabirpaul: Does this mean we are letting 2 times the limit now?

dev.c
maxkozlovsky: We allocate twice the number of ids as the maximum number of pending requests to allow for per cpu allocations. This is good for 512 CPUs.

dev.c
maxkozlovsky: See above.

dev.c
maxkozlovsky: Theoretically it can wrap around back to 0.

dev.c
maxkozlovsky: The upper bits serve as a generation count. This prevents the user space from trying to complete the same request twice. The generation count is incremented each time the id is reused. This is what += is for. There is no corresponding -=.

fuse_i.h
```diff
@@ -270,6 +267,18 @@ struct fuse_req {
...
+#define FUSE_MAX_PER_CPU_IDS 256
+
+struct ____cacheline_aligned fuse_per_cpu_ids {
+	/** number of free ids in stack */
+	u32 num_free_ids;
+
+	/** followed by list of free ids */
+	u64 free_ids[FUSE_MAX_PER_CPU_IDS];
+
+	u64 pad[7];
```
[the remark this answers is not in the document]
maxkozlovsky: removed pad

PR #63, opened 2019-12-04
remove stray printf

maxkozlovsky:


dev.c
```diff
@@ -213,9 +213,6 @@ void fuse_request_send_nowait(struct fuse_conn *fc, struct fuse_req *req)
 		len_args(req->in.numargs, (struct fuse_arg *)req->in.args);
 	spin_lock(&fc->lock);
 	if (fc->connected || fc->allow_disconnected) {
-		if (unlikely(!fc->connected)) {
-			printk(KERN_INFO "%s: Request on disconnected FC", __func__);
```
prabirpaul: This is required!
This print was how we figured out why system reboot was getting stuck and px reboot service was created.

maxkozlovsky: We don't add debug code for fixed problems.

The rules are really simple, no per request printfs for normal operation.

PR #66, opened 2019-12-16
don't reject open of user device if not connected

maxkozlovsky:
otherwise user application can fail if px is in process of restart

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -101,7 +101,7 @@ static int pxd_open(struct block_device *bdev, fmode_t mode)
 	int err = 0;
 
 	spin_lock(&fc->lock);
-	if (!fc->connected) {
+	if (!fc->connected && !fc->allow_disconnected) {
 		err = -ENXIO;
 	} else {
 		spin_lock(&pxd_dev->lock);


prabirpaul:
Open is expected to fail if px is down. It should succeed only after connection has been established.

maxkozlovsky:
User operations shouldn't be affected by temporary px state changes.

Let's say application opens device at t1, does io at t2. If px crashes and
restarts between t0 and t1, application succeeds. If px crashes before t1
and restarts after t1, application fails. If px crashes and restarts after
t1 it succeeds again. It doesn't make sense for these 3 scenarios to
produce different results because of millisecond difference in px crash and
restart time.

prabirpaul:
But there is no issue with that: 
- If px is up, open succeeds. If px is down, open fails.
Let application deal with this failure. And they already do, by retrying.
The assumption that this is a temporary state change is incorrect. Px may not come up or there could be pxd device with no px at all. Any caller could get stuck for 10min. Moreover this caller can be a scan generated from within px at startup.

maxkozlovsky:
In the driver there are two states, either px is going to come up, or it is
not going to come up which is decided by the timer after 10 minutes. If the
timer has not expired, the driver should be operating on the assumption
that px will be coming up. Letting application retry should be avoided, it
does not know if it should retry or not, how often it should retry, if the
retry is going to eventually succeed or not.

PR #70, opened 2019-12-19
fix the race where we would reset fc->connected in abort which 

maxkozlovsky:
allows user to submit new requests after abort and deadlock.

use syncrhonize_rcu() to make sure changes are seen by all threads
don't need to lock around checking connected status anymore


prabirpaul:
Wouldn't it be simpler to set fc->connected in pxd_abort_context() under fc lock. fuse_abort_conn() is taking this lock anyways.

maxkozlovsky:
Maybe simpler but not faster. rcu allows to avoid the locking while checking the value of connected in the data path.

PR #71, opened 2019-12-19
simplify abort fix; use delayed_work instead of

maxkozlovsky:
timer + schedule_work


prabirpaul:
Can you explain why the switch to workqueue is actually needed. Abort work doesn't need to sleep, a simple timer should be sufficient. Fc connection interrupt safety can handled without all these changes.

maxkozlovsky:
Workqueue is not any more complicated than a timer. The interface is almost the same. The difference is workqueue runs in normal thread context while timer runs in interrupt context. Using a timer for no reason at all requires all the code to be interrupt safe while it does not have to be.

PR #77, opened 2020-01-24
Set fastpath for valid fuse requests only

prabirpaul:
Fuse request allocation may fail, prevent null ptr access.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -429,7 +429,7 @@ void pxd_make_request_slowpath(struct request_queue *q, struct bio *bio)
 			bio->bi_vcnt, flags, get_op_flags(bio));
 
 	req = pxd_fuse_req(pxd_dev);
-	if (IS_ERR(req)) {
+	if (IS_ERR_OR_NULL(req)) {
 		bio_io_error(bio);
 		return BLK_QC_RETVAL;
 	}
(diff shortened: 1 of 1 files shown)


pxd.c
```diff
@@ -251,7 +251,6 @@ static struct fuse_req *pxd_fuse_req(struct pxd_device *pxd_dev)
 			 __func__, status);
 	}
 
-	req->fastpath = pxd_dev->fastpath;
```
sulakshm: Can we not just check if req is valid and update this condition. I was relying if fuse req is available, then its fastpath state is always in sync. Actually should check for IS_ERR_OR_NULL to be safe as below.

if (!IS_ERR_OR_NULL(req)) {
   req->fastpath = pxd_dev->fastpath;
}

And this implies not changing anywhere else.

pxd.c
```diff
@@ -489,6 +489,7 @@ static void pxd_rq_fn(struct request_queue *q)
 			continue;
 		}
 
+		req->fastpath = pxd_dev->fastpath;
```
sulakshm: The above check should be IS_ERR_OR_NULL too... and all callers of pxd_fuse_req should ensure its checked that way.

pxd.c
prabirpaul: pxd_fuse_req() just does one thing: allocate request. Which i think is the right way.
The initialization is at the caller.

sulakshm: Ok, in that case, I see only 2 places as you had fixed up for fastpath init. 
But the condition for failed fuse req allocation should still change to IS_ERR_OR_NULL.

prabirpaul: Right, done.

PR #81, opened 2020-01-28
store request directly in the queue instead of pointer

maxkozlovsky:
in preparation to map the queue into user space

### DIFF
--- a/dev.c
+++ b/dev.c
@@ -1161,8 +1161,11 @@ int fuse_restart_requests(struct fuse_conn *fc)
 	spin_lock(&fc->queue.w.lock);
 	sequence = fc->queue.w.sequence;
 	write = fc->queue.w.write;
-	if (read != write)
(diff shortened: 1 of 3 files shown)


dev.c
```diff
@@ -396,29 +385,20 @@ static ssize_t fuse_dev_do_read(struct fuse_conn *fc, struct file *file,
...
-		}
-
-		copied_this_time = fuse_copy_req_read(req, iter);
-
-		if (copied_this_time < 0) {
-			request_end(fc, req, -EIO);
-		} else {
-			copied += copied_this_time;
-			remain -= copied_this_time;
+	while (read != write && remain >= sizeof(struct rdwr_in)) {
```
jvinod: Does this also remove the optimization for zeros?

maxkozlovsky: no it's moved

PR #91, opened 2020-02-19
harden px-fuse io path

sulakshm:
1/ add magic/poison pattern to structures
2/ fix spin locks to avoid irq locks
3/ change ptr format from %p to %px

volume create, attach, detach in loop for upto 100 devices in a cycle passes cleanly after the change.


pxd_fastpath.c
```diff
@@ -624,23 +635,15 @@ static struct pxd_io_tracker* __pxd_init_block_replica(struct pxd_device *pxd_de
...
-
-	pxd_printk("pxd %p:__pxd_init_block_replica clone bio %p (blkg %p), resetting backing block dev to %p\n",
-			pxd_dev, clone_bio, clone_bio->bi_blkg, bdi);
 
+	clone_bio->bi_private = pxd_dev;
 	if (S_ISBLK(inode->i_mode)) {
 		BIO_SET_DEV(clone_bio, bdi);
+		clone_bio->bi_end_io = pxd_complete_io;
+	} else {
+		clone_bio->bi_end_io = pxd_complete_io_dummy;
```
prabirpaul: I'm not sure this is helpful. Why not just fail the op?

pxd_fastpath.c
```diff
@@ -151,7 +151,7 @@ static int fastpath_global_threadctx_init(struct thread_context *tc, int cpuid)
 	for(;i >= 0; i--) {
 		if (tc->writer[i]) kthread_stop(tc->writer[i]);
 	}
-	i = PXD_MAX_THREAD_PER_CPU;
+	i = PXD_MAX_THREAD_PER_CPU - 1;
```
prabirpaul: Why not for fail_wr?

pxd.c
```diff
@@ -81,6 +81,7 @@ static void pxd_release(struct gendisk *disk, fmode_t mode)
 	struct pxd_device *pxd_dev = disk->private_data;
 
 	spin_lock(&pxd_dev->lock);
+	BUG_ON(pxd_dev->magic != PXD_DEV_MAGIC);
```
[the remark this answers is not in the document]
prabirpaul: Please revisit all the instance.

pxd_fastpath.h
```diff
@@ -31,6 +31,8 @@ struct node_cpu_map {
 
 // Added metadata for each bio
 struct pxd_io_tracker {
+#define PXD_IOT_MAGIC (0xbeefcafe)
```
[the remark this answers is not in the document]
sulakshm: the macros are defined within the struct where they are to be used. I think this makes more sense.

pxd.c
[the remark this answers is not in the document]
sulakshm: I see, but these conditions are never ever to happen and if so, the system is hosed, a reboot is necessary.
Masking them defeats the purpose of having introduced them in the first place.
And I see px-fuse driver already has precedence:
iov_iter.c:     BUG_ON(!(direction & ITER_KVEC));
iov_iter.c:     BUG_ON(!(direction & ITER_BVEC));
dev.c:          BUG_ON(fc->num_free_ids == 0);
dev.c:          BUG_ON(fc->num_free_ids + num_free > FUSE_MAX_REQUEST_IDS);
dev.c:          BUG_ON(next_index == fc->queue->w.read);

pxd_fastpath.c
sulakshm: I dont follow your comment. If the IO is for a backed up file (i.e. file io, then bi_end_io should never be called). This field gets used only for block io. This just initializes the bio_end_io to a dummy callback, which catches any inconsistent callbacks if ever received, which should be never.

pxd_fastpath.c
```diff
@@ -1272,15 +1293,21 @@ void pxd_make_request_fastpath(struct request_queue *q, struct bio *bio)
...
-			printk("pxd device %llu re-activated, IO resumed[bio %p, wr %d]\n",
+			wait_event_interruptible_locked(pxd_dev->fp.suspend_wait, !pxd_dev->fp.suspend);
+			printk("pxd device %llu re-activated, IO resumed[bio %px, wr %d]\n",
 				pxd_dev->dev_id, bio, (bio_data_dir(bio) == WRITE));
 		}
-		spin_unlock_irq(&pxd_dev->fp.suspend_lock);
+		spin_unlock(&pxd_dev->fp.suspend_wait.lock);
+	}
+
+	if (pxd_dev->removing) {
```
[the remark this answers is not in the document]
sulakshm: This pxd dev context is going away, a reconnected px does not change that.

pxd_fastpath.c
sulakshm: because read threads are initialized first and then write threads. If there is a write thread init failure, then cleanup starts from the current index and goes back to zeroth index, then starts again to clean reads threads from the beginning. 
if read threads fail, then cleanup happens from that index and goes down to zero.
So the reinit to last index is needed only while cleaning partial write threads and all reads threads.

pxd.c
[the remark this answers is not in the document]
prabirpaul: You would end up with kernel panic with no info or a hung system which can't be logged into. (I didn't say reboot is not necessary). If you think you can debug such system then it's fine.

pxd_fastpath.c
prabirpaul: If this can never happen why setup the dummy callback at all? If it's an error condition return err to the caller.

sulakshm: to catch any issues in our fuse driver. This is not a end user error condition, this will be a px fuse driver error condition.

PR #99, opened 2020-03-14
code cleanup

maxkozlovsky:


pxd.c
```diff
@@ -89,6 +89,46 @@ static void pxd_release(struct gendisk *disk, fmode_t mode)
...
+	return 0;
+}
+
+static long pxd_ioctl_get_version(void __user *argp)
+{
+	char ver_data[64];
+	int ver_len = 0;
+
+	if (argp) {
+		ver_len = strlen(gitversion) < 64 ? strlen(gitversion) : 64;
```
sulakshm: I see that the version string now includes branch info as well, if that is causing this overflow, do you want to trim the branch and fetch only the git commit. with the current length based truncate, it may not hold the full commit id.

maxkozlovsky: I did not change any logic, this change only moves the code around.

PR #101, opened 2020-03-16
use memory mapped queue to deliver request completions

maxkozlovsky:


dev.c
```diff
@@ -164,24 +164,31 @@ static void fuse_put_unique(struct fuse_conn *fc, u64 uid)
...
-		BUG_ON(next_index == fc->queue->w.read);
+	u32 write;
+	struct rdwr_in *rdwr;
+	struct fuse_queue_cb *cb = &fc->queue->requests_cb;
+
+	spin_lock(&cb->w.lock);
+	write = cb->w.write;
+	if (write - cb->w.read >= FUSE_REQUEST_QUEUE_SIZE) {
+		cb->w.read = cb->r.read;
+		BUG_ON(write - cb->w.read >= FUSE_REQUEST_QUEUE_SIZE);
```
sulakshm: can you please elaborate the fuse_conn (w{read,write} and r{read,write}), as to who(kernel/px-storage) is supposed to update and when, so I understand the direction and definition of this usage correctly.

dev.c
```diff
@@ -164,24 +164,31 @@ static void fuse_put_unique(struct fuse_conn *fc, u64 uid)
...
-	fc->queue->requests[write].in = req->in;
-	fc->queue->requests[write].rdwr = req->pxd_rdwr_in;
-	req->sequence = fc->queue->w.sequence++;
-	fc->queue->w.write = next_index;
-	smp_store_release(&fc->queue->r.write, next_index);
-	spin_unlock(&fc->queue->w.lock);
+	rdwr = fuse_rdwr(fc, write);
+	rdwr->in = req->in;
+	rdwr->rdwr = req->pxd_rdwr_in;
+	req->sequence = cb->w.sequence++;
```
sulakshm: what are sequences for?

dev.c
```diff
@@ -232,9 +239,10 @@ void fuse_request_send_nowait(struct fuse_conn *fc, struct fuse_req *req)
 	}
 }
 
-static int request_pending(struct fuse_conn *fc)
+static bool request_pending(struct fuse_conn *fc)
 {
-	return fc->queue->r.read != fc->queue->r.write;
+	struct fuse_queue_cb *cb = &fc->queue->requests_cb;
+	return cb->r.read != cb->r.write;
```
sulakshm: this looks like the call for px-storage to check whether requests are pending unprocessed. So "r" is owned and moved by px-storage?

maxkozlovsky:
There are two unidirectional single producer single consumer queues, kernel->user space and user space->kernel. For the former, kernel is the writer, user space is the reader, for the latter user space is the writer, kernel is the reader.

Each queue has writer control block (w) updated exclusively by the writer. The writer control block used to avoid reading read index unnecessary by caching read index and batch updates to the queue by caching write index. Multiple writers synchronize with w.lock spinlock. The 'r' control block is updated both by reader and writer. During normal operation the r.write index is updated by the writer only, the r.read index is updated by reader only. During px-storage restarts the user space is not running and the kernel can update the queues however it pleases.

dev.c
maxkozlovsky: for restarts see fuse_restart_requests()

dev.c
maxkozlovsky: yes for requests queue

PR #102, opened 2020-03-20
Iouring

maxkozlovsky:


dev.c
```diff
@@ -164,24 +164,31 @@ static void fuse_put_unique(struct fuse_conn *fc, u64 uid)
...
+
 static void queue_request(struct fuse_conn *fc, struct fuse_req *req)
 {
-	u32 write, next_index;
-
-	spin_lock(&fc->queue->w.lock);
-	write = fc->queue->w.write;
-	next_index = (write + 1) & (FUSE_REQUEST_QUEUE_SIZE - 1);
-	if (fc->queue->w.read == next_index) {
-		fc->queue->w.read = fc->queue->r.read;
```
sulakshm: whats the original logic here? when a request gets queued from within the kernel, if the next index matches w.read, I assumed the queue is full. But l do not understand the next piece of logic to correlate with r.read. can you please elaborate.

dev.c
```diff
@@ -232,9 +239,10 @@ void fuse_request_send_nowait(struct fuse_conn *fc, struct fuse_req *req)
 	}
 }
 
-static int request_pending(struct fuse_conn *fc)
+static bool request_pending(struct fuse_conn *fc)
 {
-	return fc->queue->r.read != fc->queue->r.write;
+	struct fuse_queue_cb *cb = &fc->queue->requests_cb;
+	return cb->r.read != cb->r.write;
```
sulakshm: this change is fine. I see r.read and r.write as a ring pointers. So we have two independent rings, one on the kernel side receiving and informing data to px-storage (this is the "w" ring) and the other is the px-storage completing processing thats being informed through the "r" ring. Two independent rings right, but i see some correlation between "w" ring and "r" ring sometimes. havent understood that piece.

PR #104, opened 2020-03-26
fix compile, assert and code cleanup

sulakshm:

* compilation failure in centos 7.6 3.10.0-957.el7.x86_64 kernel - fixed
* IO assert - fixed
* export io distribution across thread context.
* general code cleanup.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1109,6 +1109,24 @@ static ssize_t pxd_active_show(struct device *dev,
 	return ncount;
 }
 
+// show io distribution across thread context (useful in fastpath only)
+static ssize_t pxd_distrib_show(struct device *dev,
+                     struct device_attribute *attr, char *buf)
+{
+	char *cp = buf;
+	int ncount = 0;
+	int available = PAGE_SIZE - 1;
+	int i;
+
+	for (i=0; i<num_online_cpus(); i++) {
+		size_t tmp = snprintf(cp, available, "[%d]=%d\n", i, get_thread_count(i));
+		cp += tmp;
+		available -= tmp;
+		ncount += tmp;
+	}
+
+	return ncount;
+}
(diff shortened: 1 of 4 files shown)


pxd_fastpath.c
```diff
@@ -15,7 +15,7 @@
 
 #ifdef RHEL_RELEASE_CODE
 
-#if RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(7,7)
+#if RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(7,6)
```
sulakshm: This is the reason for kernel compile failure in 3.10.0-957.el7.x86_64

pxd_fastpath.c
```diff
@@ -716,7 +769,6 @@ static void pxd_complete_io(struct bio* bio, int error)
 	__pxd_cleanup_block_io(head);
 
 	BUG_ON(pxd_dev->magic != PXD_DEV_MAGIC);
-	BUG_ON((PXD_ACTIVE(pxd_dev) == 0));
```
sulakshm: This is the IO assert reason, there may be active requests from other threads.

pxd.c
```diff
@@ -1109,6 +1109,24 @@ static ssize_t pxd_active_show(struct device *dev,
...
+// show io distribution across thread context (useful in fastpath only)
+static ssize_t pxd_distrib_show(struct device *dev,
+                     struct device_attribute *attr, char *buf)
+{
+	char *cp = buf;
+	int ncount = 0;
+	int available = PAGE_SIZE - 1;
+	int i;
+
+	for (i=0; i<num_online_cpus(); i++) {
```
prabirpaul: space around operator.

PR #110, opened 2020-04-08
set return code correctly

maxkozlovsky:


io.c
```diff
@@ -696,6 +696,7 @@ static int io_read(struct io_kiocb *req, const struct sqe_submit *s,
 	ret2 = call_read_iter(file, kiocb, &iter);
 	if (!force_nonblock || ret2 != -EAGAIN) {
 		io_rw_done(kiocb, ret2);
+		ret = 0;
```
sulakshm, line 699: This change seems not necessary, though being explicit is ok.

PR #111, opened 2020-04-08
fix duplicate open check, return ebusy on dup open

maxkozlovsky:
re-initialize volatile part of context on open

### DIFF
--- a/io.c
+++ b/io.c
@@ -698,6 +698,7 @@ static int io_read(struct io_kiocb *req, const struct sqe_submit *s,
 	ret2 = call_read_iter(file, kiocb, &iter);
 	if (!force_nonblock || ret2 != -EAGAIN) {
 		io_rw_done(kiocb, ret2);
+		ret = 0;
 	} else {
 		/*
 		 * If ->needs_lock is true, we're already in async
@@ -768,6 +769,7 @@ static int io_write(struct io_kiocb *req, const struct sqe_submit *s,
 	ret2 = call_write_iter(file, kiocb, &iter);
 	if (!force_nonblock || ret2 != -EAGAIN) {
 		io_rw_done(kiocb, ret2);
+		ret = 0;
 	} else {
 		/*
 		 * If ->needs_lock is true, we're already in async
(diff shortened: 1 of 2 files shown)


io.c
```diff
@@ -238,6 +238,8 @@ static int io_ring_ctx_init(struct io_ring_ctx *ctx)
 {
 	int i;
 
+	memset(ctx, 0, offsetof(struct io_ring_ctx, miscdev));
```
sulakshm, line 241: can we please add a comment in io.h for the io_ring_ctx, that all fields below miscdev are persisted, and above are reset every open. Keep this comment within the struct, so anyone to update it will understand the significance of new fields that go here.

io.h
```diff
@@ -132,6 +132,7 @@ struct io_ring_ctx {
 
 	struct miscdevice miscdev;
```
sulakshm, line 133: A comment right above miscdev.

io.c
```diff
@@ -696,6 +698,7 @@ static int io_read(struct io_kiocb *req, const struct sqe_submit *s,
 	ret2 = call_read_iter(file, kiocb, &iter);
 	if (!force_nonblock || ret2 != -EAGAIN) {
 		io_rw_done(kiocb, ret2);
+		ret = 0;
```
sulakshm, line 701: This does not seem needed, as "ret" anything other than zero, would have failed in the path above this call.

io.c
```diff
@@ -766,6 +769,7 @@ static int io_write(struct io_kiocb *req, const struct sqe_submit *s,
 	ret2 = call_write_iter(file, kiocb, &iter);
 	if (!force_nonblock || ret2 != -EAGAIN) {
 		io_rw_done(kiocb, ret2);
+		ret = 0;
```
sulakshm, line 772: same here.

maxkozlovsky: ret is actually -EAGAIN here. In the other case set it for consistency with this code.

io.c
maxkozlovsky: added comment

PR #113, opened 2020-04-13
malloc/free io context on open

maxkozlovsky:
having single context and checking duplicate open didn't work out well

### DIFF
--- a/io.c
+++ b/io.c
@@ -238,7 +238,7 @@ static int io_ring_ctx_init(struct io_ring_ctx *ctx)
 {
 	int i;
 
-	memset(ctx, 0, offsetof(struct io_ring_ctx, miscdev));
+	memset(ctx, 0, sizeof(*ctx));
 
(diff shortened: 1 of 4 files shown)


io.c
```diff
@@ -2349,15 +2339,15 @@ struct file_operations io_ring_fops = {
...
 
-int io_ring_register_device(char *name, struct io_ring_ctx *ctx, int context_id)
+int io_ring_register_device(struct io_dev *ctx, int context_id)
 {
 	struct miscdevice *dev = &ctx->miscdev;
 
 	ctx->context_id = context_id;
 
-	sprintf(name, "pxd/pxd-io-%d", context_id);
+	sprintf(ctx->name, "pxd/pxd-io-%d", context_id);
```
sulakshm, line 2348: ctx->name is defined 256 bytes, but this name format will not go more than 22 bytes approx. Maybe can be reduced to 64 chars atleast... even with sprintf usage it should suffice isn't it?

PR #116, opened 2020-04-28
add reader spinlock for user space

maxkozlovsky:


fuse_i.h
```diff
@@ -132,7 +133,8 @@ struct alignas(64) fuse_queue_writer {
 struct alignas(64) fuse_queue_reader {
 	std::atomic<uint32_t> read;	/** read index updated by reader */
 	std::atomic<uint32_t> write;	/** write index updated by writer */
-	uint64_t pad_2[7];
+	px::spinlock lock;
+	uint64_t pad_2[6];
```
sulakshm, line 137: This struct definition looks scary... how are we ensuring that fuse_queue_reader|writer defined by the kernel and the userspace are of same size? And embedding complex objects and assuming they remain same size across different versions is scary to me... 
Could we atleast hardcode the max size of the struct and ensure during compile time they remain the same between the kernel and the userspace?

maxkozlovsky: It's not scary at all, the tests will break if the structures go out of sync. Since we run the tests before every merge there is zero possibility for this to happen.

PR #118, opened 2020-05-06
fix leak; cleanup the state each time through the loop

maxkozlovsky:


sulakshm:
Can you also please fix the formatting in this file?

maxkozlovsky:
> Can you also please fix the formatting in this file?

What is wrong with the formatting? Looks perfect to me

PR #119, opened 2020-05-08
add option to force async submission for direct writes, because btrfs

maxkozlovsky:


maxkozlovsky:
> I do not understand this change, can you please elaborate?
> All file IO requests are submitted as one of below.
> 
> ```
> enum class Command {
>         nop = IORING_OP_NOP,
>         readv = IORING_OP_READV,
>         writev = IORING_OP_WRITEV,
> };
> ```
> 
> Now, these are run inline in the kernel context from within io_submit_sqe, irrespective of whether O_DIRECT is specified.
> Please clarify more on this change.

If O_DIRECT is set, io_submit will try to queue request non-blocking and switch to async if that does not work, btrfs does this poorly and blocks the request anyway. The change is to bypass non-blocking submission if the flag is set to avoid blocking in btrfs code.

maxkozlovsky:
> Line 524-528: should the set not be reversed in order.
> Also, one should check FMODE_NOWAIT is supported by the filesystem before forcing NOWAIT flag to be set. This can be done during file open and helps avoid pitfalls.

The code looks correct to me.

sulakshm:
> I do not understand this change, can you please elaborate?
> All file IO requests are submitted as one of below.
> enum class Command {
>         nop = IORING_OP_NOP,
>         readv = IORING_OP_READV,
>         writev = IORING_OP_WRITEV,
> };
> 
> Now, these are run inline in the kernel context from within io_submit_sqe, irrespective of whether O_DIRECT is specified.
> Please clarify more on this change.
> 
> If O_DIRECT is set, io_submit will try to queue request non-blocking and switch to async if that does not work, btrfs does this poorly and blocks the request anyway. The change is to bypass non-blocking submission if the flag is set to avoid blocking in btrfs code.

The first IO submission from io_ring_submit() always submits in force_nonblock = true, irrespective of whether file is opened in O_DIRECT and if the response is -EAGAIN, then it is reenqueued through async context. Per your note above, btrfs O_DIRECT blocks anyways, and this change is supposed to force that IO request through the async path. But I do not see that change.
Now adding this new flag, does not change whats described above.
And the only change in PR is to hint for read ahead for async write IO.
So it still seems the change is incomplete. Please clarify.

maxkozlovsky:
If the write is not direct, the code always schedules the write through the work queue. The change adds an option to force scheduling direct writes through the work queue bypassing non-blocking attempt. Here is the user space part: https://github.com/portworx/porx/pull/5766.

PR #120, opened 2020-05-09
optimize IO scheduling path 

sulakshm:
With the below changes addressed the high latency while in fastpath.
Results are tabulated here under optimized heading:
https://portworx.atlassian.net/wiki/spaces/PC/pages/899121235/Performance+study+on+fastpath+volumes
a) removed the queueing logic and if request if for block device requests can go through direct without any queueing.
b) for file based backing device, needs to be queued to avoid deadlocks.
c) optimize and remove all locks in the regular IO path.
d) use work queue instead for IO queueing and processing file IO.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -711,7 +711,6 @@ static void pxd_free_disk(struct pxd_device *pxd_dev)
 	if (!disk)
 		return;
 
-	pxd_fastpath_cleanup(pxd_dev);
 	pxd_dev->disk = NULL;
 	if (disk->flags & GENHD_FL_UP) {
 		del_gendisk(disk);
@@ -875,9 +874,9 @@ ssize_t pxd_remove(struct fuse_conn *fc, struct pxd_remove_out *remove)
 	}
 
 	spin_unlock(&pxd_dev->lock);
-	pxd_fastpath_cleanup(pxd_dev);
 
 	device_unregister(&pxd_dev->dev);
+	pxd_fastpath_cleanup(pxd_dev);
 
 	module_put(THIS_MODULE);
 
--- a/pxd.c
+++ b/pxd.c
@@ -1144,24 +1144,6 @@ static ssize_t pxd_active_show(struct device *dev,
 	return ncount;
 }
 
-// show io distribution across thread context (useful in fastpath only)
-static ssize_t pxd_distrib_show(struct device *dev,
-                     struct device_attribute *attr, char *buf)
-{
-	char *cp = buf;
-	int ncount = 0;
-	int available = PAGE_SIZE - 1;
-	int i;
-
-	for (i = 0; i < num_online_cpus(); i++) {
-		size_t tmp = snprintf(cp, available, "[%d]=%d\n", i, get_thread_count(i));
-		cp += tmp;
-		available -= tmp;
-		ncount += tmp;
-	}
-
-	return ncount;
-}
 static ssize_t pxd_sync_show(struct device *dev,
                      struct device_attribute *attr, char *buf)
 {
@@ -1379,7 +1361,6 @@ static DEVICE_ATTR(major, S_IRUGO, pxd_major_show, NULL);
 static DEVICE_ATTR(minor, S_IRUGO, pxd_minor_show, NULL);
 static DEVICE_ATTR(timeout, S_IRUGO|S_IWUSR, pxd_timeout_show, pxd_timeout_store);
 static DEVICE_ATTR(active, S_IRUGO, pxd_active_show, NULL);
-static DEVICE_ATTR(distrib, S_IRUGO, pxd_distrib_show, NULL);
 static DEVICE_ATTR(sync, S_IRUGO|S_IWUSR, pxd_sync_show, pxd_sync_store);
 static DEVICE_ATTR(congested, S_IRUGO|S_IWUSR, pxd_congestion_show, pxd_congestion_set);
 static DEVICE_ATTR(writesegment, S_IRUGO|S_IWUSR, pxd_wrsegment_show, pxd_wrsegment_store);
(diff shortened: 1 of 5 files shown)


pxd_fastpath.c
```diff
@@ -1023,14 +799,23 @@ static int __do_bio_filebacked(struct pxd_device *pxd_dev, struct pxd_io_tracker
 
 	return ret;
 }
-
 #endif
 
-static
-int pxd_handle_io(struct thread_context *tc, struct pxd_io_tracker *head, int dir)
+static void pxd_process_fileio(struct work_struct *wi)
```
sulakshm, line 804: file IO will use this routine to complete IO to the backing file. This is run through a work queue.

pxd_fastpath.c
```diff
@@ -889,13 +672,6 @@ static int __do_bio_filebacked(struct pxd_device *pxd_dev, struct pxd_io_tracker
 
 	BUG_ON(pxd_dev->magic != PXD_DEV_MAGIC);
 	BUG_ON(iot->magic != PXD_IOT_MAGIC);
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(4,14,0)
```
sulakshm, line 892: This was duplicated. Should be initialized only once at pxd_process_io.

pxd_fastpath.c
```diff
@@ -1023,14 +799,23 @@ static int __do_bio_filebacked(struct pxd_device *pxd_dev, struct pxd_io_tracker
...
+{
+	struct pxd_io_tracker *iot = container_of(wi, struct pxd_io_tracker, wi);
+	struct pxd_device *pxd_dev = iot->pxd_dev;
+
+	BUG_ON(iot->magic != PXD_IOT_MAGIC);
+	BUG_ON(pxd_dev->magic != PXD_DEV_MAGIC);
+	__do_bio_filebacked(pxd_dev, iot);
+}
+
+static void pxd_process_io(struct pxd_io_tracker *head)
```
sulakshm, line 814: IO core processing is now direct instead of from a queue earlier. This is the main routine.

pxd_fastpath.c
```diff
@@ -1545,7 +1308,7 @@ void pxd_fastpath_adjust_limits(struct pxd_device *pxd_dev, struct request_queue
 		if (!bdev || IS_ERR(bdev)) {
 			printk(KERN_ERR"pxd device %llu: backing block device lookup for path %s failed %ld\n",
 				pxd_dev->dev_id, pxd_dev->fp.device_path[i], PTR_ERR(bdev));
-			continue;
+			goto out;
```
sulakshm, line 1311: If this fails on a block device for any reason, then the queue parameters between px volume and the backing device may not be compatible and may result in failed IO. Hence failing to native path.

pxd.c
```diff
@@ -1144,24 +1143,6 @@ static ssize_t pxd_active_show(struct device *dev,
 	return ncount;
 }
 
-// show io distribution across thread context (useful in fastpath only)
```
sulakshm, line 1147: removing unused.

pxd_fastpath.c
```diff
@@ -107,36 +108,29 @@ void _generic_start_io_acct(struct request_queue *q, int rw,
...
-#define pxd_get_readio(tc)   pxd_get_io(tc, READ)
 
 // congestion callback from kernel writeback module
 int pxd_device_congested(void *data, int bits)
 {
 	struct pxd_device *pxd_dev = data;
 	int ncount = PXD_ACTIVE(pxd_dev);
+	int cpu = get_cpu();
+	struct pcpu_fpstate *statep = per_cpu_ptr(pxd_dev->fp.state, cpu);
+	int suspend = READ_ONCE(statep->suspend);
```
prabirpaul, line 127: Why is it ok to access the percpu without lock here?

pxd_fastpath.c
```diff
@@ -1513,8 +1255,29 @@ void pxd_make_request_fastpath(struct request_queue *q, struct bio *bio)
 		return BLK_QC_RETVAL;
 	}
 
-	tc = get_thread_context(rw);
-	pxd_add_io(tc, head, rw);
+{
+	// If IO suspended, then hang IO onto the suspend wait queue
+	int cpu = get_cpu();
```
prabirpaul, line 1260: you can just make a single call to get the local percpu rt?

pxd_fastpath.c
```diff
@@ -1198,18 +906,52 @@ static void pxd_suspend_io(struct pxd_device *pxd_dev)
...
-	spin_lock(&pxd_dev->fp.suspend_wait.lock);
-	pxd_dev->fp.suspend--;
-	wakeup = (pxd_dev->fp.suspend == 0);
-	spin_unlock(&pxd_dev->fp.suspend_wait.lock);
-
+	int cpu, new = 0, old = 0;
+
+	BUG_ON(!pxd_dev->fp.state);
+	spin_lock(&pxd_dev->fp.suspend_lock);
+	for_each_online_cpu(cpu) {
```
prabirpaul, line 915: If both susp/resume and going to iterate over all cpus, it needs locked lookup. It also means local cpu reads need lock. Why use percpu at all?
If u are going to use percpu, doesnt it make more sense to partition the fp.suspend_queue, so list_add is done percpu?

pxd_fastpath.c
sulakshm: The primary reason per cpu got introduced is in the IO path, in make_request, when the IO is processed, if its not suspended, is processed straight. If suspended, then it is confirmed under lock and enqueued. So the common operation is always without lock, because suspending IO happens only on congestion, failover etc.

the lock (suspend_lock) is needed to ensure suspend queuing is consistent with suspend request. Here one relies only on the current state, so suspend_lock is not needed. This queueing in IO path is going away in failover patch next cycle.

pxd_fastpath.c
sulakshm: get_cpu ensures preemption is disabled, so context is still correct.

pxd_fastpath.c
sulakshm: Only the control operation (i.e. a suspend or a resume call) needs to ensure all cpu have the same state. 
Regular IO path through make_request() just checks current state only within its local cpu ptr.

PR #122, opened 2020-05-11
discard/syncfs support on io uring

sulakshm:


io.c
```diff
@@ -784,6 +784,87 @@ static int io_write(struct io_kiocb *req, const struct sqe_submit *s,
...
+	int ret = -EINVAL;
+	loff_t pos = s->pos;
+	size_t bytes = s->len;
+	
+	/* discard always requires a blocking context */
+	if (force_nonblock)
+		return -EAGAIN;
+
+	if (unlikely(!(req->file->f_mode & FMODE_WRITE)))
+		return -EBADF;
```
maxkozlovsky, line 802: EBADF does not look like the right return code here

io.c
```diff
@@ -784,6 +784,87 @@ static int io_write(struct io_kiocb *req, const struct sqe_submit *s,
...
+	} else  if (S_ISBLK(inode->i_mode)) {
+		struct block_device *bdev = I_BDEV(inode);
+		sector_t start_sector = (pos >> SECTOR_SHIFT);
+		sector_t nsectors = sector_div(bytes, SECTOR_SIZE);
+
+		// santize byte range for sector granularity
+		if ((pos & (SECTOR_SIZE-1)) || (nsectors & (SECTOR_SIZE-1))) {
+			return -EINVAL;
+		}
+		ret = blkdev_issue_discard(bdev, start_sector, nsectors, GFP_KERNEL, 0);
```
maxkozlovsky, line 822: I think in 4.x+ kernels the block devices support fallocate. uring is supported from 4.15, so you might be able to simply call fallocate all the time.

io.c
```diff
@@ -784,6 +784,87 @@ static int io_write(struct io_kiocb *req, const struct sqe_submit *s,
...
+	} else {
+		return -EINVAL;
+	}
+
+	// only for success do the below, as its done through the wq caller context
+	if (!req) {
```
maxkozlovsky, line 828: if (!ret)

io.c
```diff
@@ -784,6 +784,87 @@ static int io_write(struct io_kiocb *req, const struct sqe_submit *s,
...
+		up_read(&sb->s_umount);
+	} else if (S_ISBLK(inode->i_mode)) {
+		struct block_device *bdev = I_BDEV(inode);
+		ret = blkdev_issue_flush(bdev, GFP_KERNEL, NULL);
+	} else {
+		return -EOPNOTSUPP;
+	}
+
+	// only for success do the below, as its done through the wq caller context
+	if (!req) {
```
maxkozlovsky, line 860: if (!ret)

io.c
sulakshm: fixed, thanks,

PR #126, opened 2020-06-03
Resize ioctl

prabirpaul:
This is to allow resize operation to be called from outside of px-storage. When resizing, revalidate_disk could be a long running operation. If px dies while the op is in progress, px-storage ends up in D state.


pxd.c
```diff
@@ -146,6 +146,32 @@ static long pxd_ioctl_init(struct file *file, void __user *argp)
...
+{
+	struct pxd_context *ctx = NULL;
+	struct pxd_update_size_out update_args;
+	long ret = 0;
+
+	if (copy_from_user(&update_args, argp, sizeof(update_args))) {
+		return -EFAULT;
+	}
+
+	if (!ctx || ctx->id >= pxd_num_contexts_exported) {
```
sulakshm, line 159: This condition is incorrect, ctx is initialized only below

prabirpaul: merge error, fixed now.

PR #127, opened 2020-06-08
address use after free (fuse_req)

sulakshm:
Running good so far:
https://jenkins.portworx.co/job/DEV/job/Porx-05/213

The crash only applies if the block device is registered as a blkmq device.

### DIFF
--- a/dev.c
+++ b/dev.c
@@ -210,6 +210,7 @@ static void request_end(struct fuse_conn *fc, struct fuse_req *req,
 	int status)
 {
 	u64 uid = req->in.unique;
+	bool using_blkque = req->pxd_dev->using_blkque;
 	if (req->end)
 		req->end(fc, req, status);
 	fuse_put_unique(fc, uid);
@@ -217,7 +218,7 @@ static void request_end(struct fuse_conn *fc, struct fuse_req *req,
 #ifndef __PX_BLKMQ__
 	fuse_request_free(req);
 #else
-	if (!req->pxd_dev->using_blkque) fuse_request_free(req);
+	if (!using_blkque) fuse_request_free(req);
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


dev.c
```diff
@@ -210,14 +210,17 @@ static void request_end(struct fuse_conn *fc, struct fuse_req *req,
 	int status)
 {
 	u64 uid = req->in.unique;
+#ifdef __PX_BLKMQ__
+	bool using_blkque = req->pxd_dev->using_blkque;
+#endif
 	if (req->end)
 		req->end(fc, req, status);
```
sulakshm, line 217: after this call if blkmq block device was used, fuse req (req) is no longer available. crash below.

pxd.c
```diff
@@ -305,24 +305,24 @@ static void pxd_process_write_reply(struct fuse_conn *fc, struct fuse_req *req,
 static void pxd_process_read_reply_q(struct fuse_conn *fc, struct fuse_req *req,
 	int status)
 {
+	pxd_request_complete(fc, req);
 #ifndef __PX_BLKMQ__
 	blk_end_request(req->rq, status, blk_rq_bytes(req->rq));
 #else
 	blk_mq_end_request(req->rq, errno_to_blk_status(status));
```
sulakshm, line 312: this call frees up bio and the fuse req together.

pxd.c
```diff
@@ -1012,8 +1012,8 @@ ssize_t pxd_remove(struct fuse_conn *fc, struct pxd_remove_out *remove)
 
 	spin_unlock(&pxd_dev->lock);
 
-	device_unregister(&pxd_dev->dev);
 	pxd_fastpath_cleanup(pxd_dev);
+	device_unregister(&pxd_dev->dev);
```
sulakshm, line 1016: device_unregister may free up pxd_dev, moving fastpath cleanup before.

PR #128, opened 2020-06-18
Resize ioctl (#126)

prabirpaul:
Cherry-pick from master.
px-fuse master has too many other changes, hence branching it for v2.6.


pxd.h
```diff
@@ -146,11 +147,12 @@ struct pxd_read_data_out {
 };
 
 /**
- * PXD_UPDATE_SIZE request from user space
+ * PXD_UPDATE_SIZE ioctl from user space
  */
-struct pxd_update_size_out {
+struct pxd_update_size {
```
sulakshm, line 152: have you made sure all references to pxd_update_size_out are taken care? Looks like pxdriver still has it.

PR #132, opened 2020-07-01
Allow greater than 300sec timeout values

prabirpaul:
We need to be able to set the timeout to a higher value. This is required when a maintenance operation is undergoing on node which would keep it down for more than 10 minutes.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -36,7 +36,8 @@
 #endif
 
 #define PXD_TIMER_SECS_MIN 30
-#define PXD_TIMER_SECS_MAX 600
+#define PXD_TIMER_SECS_DEFAULT 600
+#define PXD_TIMER_SECS_MAX (U32_MAX)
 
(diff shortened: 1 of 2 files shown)


sulakshm:
The changes are okay, but during maintenance why would we want to hold IO and not fail immediately. Already 10 mins is a long time isn't it? All applications(pods) running on the node would have been rescheduled because node is in maintenance.
So what purpose does not failing IO and that too this long really serve?

prabirpaul:
We aren't changing the default timeout. It is still 10min. Maintenance could be external to the node. For example px could be down because of etcd unavailable. And that could take more than 10mins to recover. The change adds the ability prevent the mountpoint turning readonly during this window. This is a specific ask from customer.

sulakshm:
By giving the customer a knob to increase timeout values? I am not too sure i understand, but the changes are harmless though.

PR #133, opened 2020-07-01
Ensure no failover if IO to local replica

sulakshm:
As replica paths are attached during kernel device add, also specify if the replica is local.
If the device attached is local and is a single replica, then do not failover during IO errors.

With fastpath disabled:
https://jenkins.portworx.co/job/DEV/job/Porx-05/247

With fastpath enabled:
https://jenkins.portworx.co/job/DEV/job/Porx-03/286

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -874,7 +874,7 @@ struct pxd_device* find_pxd_device(struct pxd_context *ctx, uint64_t dev_id)
 	return pxd_dev;
 }
 
-static int __pxd_update_path(struct pxd_device *pxd_dev, bool can_failover,
+static int __pxd_update_path(struct pxd_device *pxd_dev,
 		struct pxd_update_path_out *update_path);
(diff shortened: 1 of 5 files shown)


pxd.c
```diff
@@ -874,7 +874,8 @@ struct pxd_device* find_pxd_device(struct pxd_context *ctx, uint64_t dev_id)
 	return pxd_dev;
 }
 
-static int __pxd_update_path(struct pxd_device *pxd_dev, struct pxd_update_path_out *update_path);
+static int __pxd_update_path(struct pxd_device *pxd_dev, bool can_failover,
```
prabirpaul, line 877: can_failover can be embedded in pxd_update_path_out just like pxd_add_ext_out.

sulakshm: yes can be embedded and is more appropriate. revised.

PR #139, opened 2020-07-25
non block mq device does not handle queue limits

sulakshm:
Fastpath device registers as a non block mq device, and it does not handle queue limits.
This makes received IO to be larged than PXD_MAX_IO (1MB) and crashes the state machine for both read write and discards.
This change checks and handles the incoming request by splitting requests as needed.

Below were the affected unit tests that pass now in fastpath mode.
```
root@bionic:src/gdfs# ./t --fastpath --gtest_filter=MultiNodeTest.restart_after_unmount:MultiNodeTest.discard:MultiNodeTest.try_detach_during_write
...
rmmod: ERROR: Module px is in use
[       OK ] MultiNodeTest.discard (7231 ms)
[----------] 3 tests from MultiNodeTest (38928 ms total)

[----------] Global test environment tear-down
[==========] 3 tests from 1 test suite ran. (38930 ms total)
[  PASSED  ] 3 tests.
```

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1579,6 +1579,14 @@ static ssize_t pxd_debug_store(struct device *dev,
 		printk("dev:%llu - IO fast path switch\n", pxd_dev->dev_id);
 		pxd_switch_fastpath(pxd_dev);
 		break;
+	case 'S': /* app suspend */
+		printk("dev:%llu - requesting IO suspend\n", pxd_dev->dev_id);
+		pxd_request_suspend(pxd_dev, false);
+		break;
+	case 'R': /* app resume */
+		printk("dev:%llu - requesting IO resume\n", pxd_dev->dev_id);
+		pxd_request_resume(pxd_dev);
+		break;
 	default:
 		/* no action */
 		printk("dev:%llu - no action for %c\n", pxd_dev->dev_id, buf[0]);
--- a/pxd.c
+++ b/pxd.c
@@ -1124,6 +1124,7 @@ ssize_t pxd_read_init(struct fuse_conn *fc, struct iov_iter *iter)
 		id.dev_id = pxd_dev->dev_id;
 		id.local_minor = pxd_dev->minor;
 		id.fastpath = 0;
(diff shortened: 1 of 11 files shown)


pxd.c
```diff
@@ -406,9 +407,57 @@ static void pxd_req_misc(struct fuse_req *req, uint32_t size, uint64_t off,
 #endif
 }
 
-static void pxd_read_request(struct fuse_req *req, uint32_t size, uint64_t off,
+/*
+ * when block device is registered in non blk mq mode, device limits are not
+ * honoured. Handle it appropriately.
+ */
+static
+int pxd_handle_device_limits(struct fuse_req *req, uint32_t *size, uint64_t *off,
```
prabirpaul, line 415: Isn't pxd_fastpath_adjust_limits() supposed to take care of the q limits?

sulakshm: Originally we had bio too large error message (refer: https://access.redhat.com/solutions/3613951), that was fixed by introducing these to adjust limits based on actual targets. 
In this case, even though limits were set, under some scenarios, block device drivers requests were over the limits for read/write and even discard. This caused px-storage to assert at various places. Maybe the issue is kernel version dependent, but the fix is generic.

pxd.c
```diff
@@ -459,45 +537,48 @@ static void pxd_write_same_request(struct fuse_req *req, uint32_t size, uint64_t
...
 	pxd_req_misc(req, size, off, minor, flags);
+	return 0;
 }
 
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(4,8,0) || defined(REQ_PREFLUSH)
 static int pxd_request(struct fuse_req *req, uint32_t size, uint64_t off,
 			uint32_t minor, uint32_t op, uint32_t flags)
 {
+	int rc;
 	trace_pxd_request(req->in.unique, size, off, minor, flags);
```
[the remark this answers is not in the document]
sulakshm: Yes but it will involve decoding the type of request under various version checks, which is already done within pxd_request. So keeping it as is.

PR #140, opened 2020-07-28
bio_split signature mismatch on 3.10 kernels

sulakshm:
fix the path with kernel version checks.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -432,13 +432,16 @@ int pxd_handle_device_limits(struct fuse_req *req, uint32_t *size, uint64_t *off
 
 	max_sectors = blk_queue_get_max_sectors(q, op);
 	while (rq_sectors > max_sectors) {
+		struct bio *b;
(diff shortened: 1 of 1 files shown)


pxd.c
```diff
@@ -434,8 +434,12 @@ int pxd_handle_device_limits(struct fuse_req *req, uint32_t *size, uint64_t *off
 	while (rq_sectors > max_sectors) {
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(4,18,0)
 		struct bio *b = bio_split(req->bio, max_sectors, GFP_NOIO, &fs_bio_set);
-#else
+#elif LINUX_VERSION_CODE >= KERNEL_VERSION(3,14,0)
 		struct bio *b = bio_split(req->bio, max_sectors, GFP_NOIO, fs_bio_set);
+#else
+		// This issue has so far been seen only with 4.20 and 5.x kernels
```
prabirpaul, line 440: Should add a ratelimited alert, so that we know why the io aborted.

sulakshm: done

PR #144, opened 2020-08-12
fastpath support IO fallback

sulakshm:
Extend px-fuse driver to support the following in fastpath.
1/ fallback - while fastpath capable devices are in nativepath,  a fallback request shall promote the volume to fastpath.
When volume is in native path, on fallback request, volume IO is suspend, and a fallback request gets submitted to userspace, which refreshes the SAN path and reattaches volume spec for fastpath, and re-enables IO.

2/ failover - reworked to initiate failover explicitly under all cases. When a fastpath volume receives IO error, then volume is suspended for new IO, failed requests are queued, and a failover request gets initiated to userspace, which tears down SAN path, and reattaches volumes in native path, and re-enables IO.

### DIFF
--- a/dev.c
+++ b/dev.c
@@ -237,12 +237,10 @@ void fuse_request_send_nowait(struct fuse_conn *fc, struct fuse_req *req, bool f
 
 	if (force) {
 		queue_request(fc, req);
-		rcu_read_unlock();
-
 		if (fc->connected || fc->allow_disconnected) {
 			fuse_conn_wakeup(fc);
 		}
-
+		rcu_read_unlock();
 	} else if (fc->connected || fc->allow_disconnected) {
 		queue_request(fc, req);
 		rcu_read_unlock();
--- a/dev.c
+++ b/dev.c
@@ -210,6 +210,9 @@ static void request_end(struct fuse_conn *fc, struct fuse_req *req,
 	int status)
 {
 	u64 uid = req->in.unique;
+#ifdef __PX_BLKMQ__
+	bool using_blkque = req->pxd_dev->using_blkque;
+#endif
 	if (req->end)
 		req->end(fc, req, status);
 	fuse_put_unique(fc, uid);
@@ -217,7 +220,7 @@ static void request_end(struct fuse_conn *fc, struct fuse_req *req,
 #ifndef __PX_BLKMQ__
 	fuse_request_free(req);
 #else
-	if (!req->pxd_dev->using_blkque) fuse_request_free(req);
+	if (!using_blkque) fuse_request_free(req);
 #endif
 }
 
--- a/dev.c
+++ b/dev.c
@@ -827,9 +827,8 @@ static int fuse_notify_ioswitch_event(struct fuse_conn *conn, unsigned int size,
                return -EINVAL;
        }
 
-       if (failover) return pxd_request_failover(pxd_dev);
-
-       return pxd_request_fallback(pxd_dev);
+       return pxd_request_ioswitch(pxd_dev,
+                failover ? PXD_FAILOVER_TO_USERSPACE : PXD_FALLBACK_TO_KERNEL);
 }
 
 static int fuse_notify(struct fuse_conn *fc, enum fuse_notify_code code,
--- a/dev.c
+++ b/dev.c
@@ -801,6 +801,27 @@ static int fuse_notify_resume(struct fuse_conn *conn, unsigned int size,
 	return pxd_request_resume(pxd_dev);
 }
 
+static int fuse_notify_fallback(struct fuse_conn *conn, unsigned int size,
+               struct iov_iter *iter) {
+       struct pxd_context *ctx = container_of(conn, struct pxd_context, fc);
+       struct pxd_fallback req;
+       size_t len = sizeof(req);
+       struct pxd_device *pxd_dev;
+
+       if (copy_from_iter(&req, len, iter) != len) {
+               printk(KERN_ERR "%s: can't copy arg\n", __func__);
+               return -EFAULT;
+       }
+
+       pxd_dev = find_pxd_device(ctx, req.dev_id);
+       if (!pxd_dev) {
+               printk(KERN_ERR "device %llu not found\n", req.dev_id);
+               return -EINVAL;
+       }
+
+       return pxd_request_fallback(pxd_dev);
+}
+
 static int fuse_notify(struct fuse_conn *fc, enum fuse_notify_code code,
 		       unsigned int size, struct iov_iter *iter)
 {
@@ -825,6 +846,8 @@ static int fuse_notify(struct fuse_conn *fc, enum fuse_notify_code code,
 		return fuse_notify_suspend(fc, size, iter);
 	case PXD_RESUME:
 		return fuse_notify_resume(fc, size, iter);
+    case PXD_FALLBACK:
+        return fuse_notify_fallback(fc, size, iter);
 	default:
 		return -EINVAL;
 	}
--- a/dev.c
+++ b/dev.c
@@ -224,7 +224,7 @@ static void request_end(struct fuse_conn *fc, struct fuse_req *req,
 #endif
 }
 
-void fuse_request_send_nowait(struct fuse_conn *fc, struct fuse_req *req)
+void fuse_request_send_nowait(struct fuse_conn *fc, struct fuse_req *req, bool force)
 {
 	req->in.unique = fuse_get_unique(fc);
 	fc->request_map[req->in.unique & (FUSE_MAX_REQUEST_IDS - 1)] = req;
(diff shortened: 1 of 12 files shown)


pxd_fastpath.c
```diff
@@ -1404,3 +1455,53 @@ int pxd_switch_nativepath(struct pxd_device* pxd_dev)
...
+
+int __pxd_reissuefailQ(struct pxd_device *pxd_dev, int status)
+{
+	if (!status) {
+		while (!list_empty(&pxd_dev->fp.failQ)) {
+			struct pxd_io_tracker *head = list_first_entry(&pxd_dev->fp.failQ, struct pxd_io_tracker, item);
+			BUG_ON(head->magic != PXD_IOT_MAGIC);
+			list_del(&head->item);
+			if (!pxd_dev->connected || pxd_dev->removing) {
+				printk_ratelimited(KERN_ERR"%s: pxd%llu: px is disconnected, failing IO.\n", __func__, pxd_dev->dev_id);
```
prabirpaul, line 1475: This would abort IOs immediately if px is down. IO abort should happen only if px is down for more than 10 mins.

pxd_fastpath.c
```diff
@@ -912,6 +926,41 @@ bool pxd_sync_work_pending(struct pxd_device *pxd_dev)
...
+       }
+
+       // IO path blocked, a future path refresh will take it to native path
+       // enqueue a failover request to userspace on this device.
+       printk("device %llu initiated failover\n", pxd_dev->dev_id);
+       return pxd_initiate_failover(pxd_dev);
+}
+
+// external request to initiate fallback on fastpath device
+int pxd_request_fallback(struct pxd_device *pxd_dev)
```
prabirpaul, line 948: pxd_request_failover/pxd_request_fallback can be collapsed into one.

pxd_fastpath.c
```diff
@@ -415,33 +417,43 @@ static void pxd_io_failover(struct work_struct *ws)
...
 		disableFastPath(pxd_dev, true);
-	}
-
-	if (!pxd_dev->connected || pxd_dev->removing) {
-		printk_ratelimited(KERN_ERR"%s: pxd%llu: px is disconnected, failing IO.\n", __func__, pxd_dev->dev_id);
-		BIO_ENDIO(head->orig, -ENXIO);
-	} else {
-		// switch to native path
+		rc = pxd_initiate_failover(pxd_dev);
+		// If userspace cannot be informed of a failover event, force abort all IO.
```
prabirpaul, line 442: Here if px is down temporarily during switch, all IOs will abort. Is this what u intend?

pxd.c
```diff
@@ -631,6 +631,104 @@ static inline unsigned int get_op_flags(struct bio *bio)
...
+	if (req->in.opcode == PXD_FAILOVER) {
+		// if the status is successful, then reissue IO to userspace
+		// else fail IO to complete.
+		spin_lock(&pxd_dev->fp.fail_lock);
+		__pxd_reissuefailQ(pxd_dev, status);
+		pxd_dev->fp.active_failover = PXD_FP_FAILOVER_NONE;
+		spin_unlock(&pxd_dev->fp.fail_lock);
+	}
+
+	// reopen the suspended device
```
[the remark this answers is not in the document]
sulakshm: further down... currently IOs that have errored in fastpath try to failover, and based on the status are reissued to userspace or immediately failed. I should probably extend the abort IO path on timer expiry to cleanup these IO... will revise.

pxd.c
```diff
@@ -631,6 +631,104 @@ static inline unsigned int get_op_flags(struct bio *bio)
...
+	if (pxd_dev->using_blkque) {
+		return -EINVAL;
+	}
+
+	if (atomic_cmpxchg(&pxd_dev->switch_active, 0, 1) != 0) {
+		return -EBUSY;
+	}
+
+	// if already suspended, then out again.
+	if (pxd_dev->fp.app_suspend) {
```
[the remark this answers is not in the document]
sulakshm: yes, will consolidate. actually suspend/resume (clone), failover (io failed in fastpath)/fallback(auto on timer) and node up/down all race... becoming complex now.

pxd_fastpath.c
sulakshm: correct. will revise.

pxd_fastpath.c
```diff
@@ -1404,3 +1455,53 @@ int pxd_switch_nativepath(struct pxd_device* pxd_dev)
 	}
 	return 0;
 }
+
+/// handle io path switch events and io reroute on failures
+/// functions prefixed with ___xxx need to take fail_lock
```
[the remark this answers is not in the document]
sulakshm: comment needs revision, function with ___xxxx are naked functions and caller should take locks.

pxd_fastpath.c
sulakshm: yes failure IO path needs rework to hang off abort IO timeout.... will revise.

sulakshm (commented):
thanks, good feedback. Will revise.

pxd.c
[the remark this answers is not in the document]
sulakshm: fastpath IO tries to initiate failover to userspace on any IO failure. If this fails, then IO is failed to userspace immediately.
failover initiation is through an inline IO request through the fuse path. So it can be enqueued even when px is down.

pxd.c
```diff
@@ -631,6 +631,104 @@ static inline unsigned int get_op_flags(struct bio *bio)
...
+
+	rc = pxd_initiate_ioswitch(pxd_dev, PXD_FAILOVER);
+	if (rc) {
+		pxd_request_resume(pxd_dev);
+	}
+
+	return rc;
+}
+
+int pxd_initiate_fallback(struct pxd_device *pxd_dev)
```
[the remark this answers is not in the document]
sulakshm: there are small differences in the args... the guts of the logic though is already seperated in pxd_initiate_ioswitch earlier.

pxd_fastpath.c
sulakshm: IO will be enqueued to userspace when failover request completes. But if failover request itself fails, then the IOs fail immediately to end user.

pxd_fastpath.c
sulakshm: done, introduced pxd_request_ioswitch()

sulakshm (commented):
addressed review comments

pxd_fastpath.c
```diff
@@ -995,19 +1051,32 @@ void pxd_suspend_io(struct pxd_device *pxd_dev)
...
-	pxd_dev->fp.app_suspend = false;
 	printk(KERN_NOTICE"device %llu resumed IO from userspace\n", pxd_dev->dev_id);
 	return 0;
 }
 
+// external request to resume IO on fastpath device
+int pxd_request_resume(struct pxd_device *pxd_dev)
+{
+	int rc;
+	if (atomic_read(&pxd_dev->fp.app_suspend) == 0) {
```
prabirpaul, line 1069: Don't u need a CAS here as well? or is it ok to let parallel calls to pxd_request_resume_internal through?

pxd_fastpath.c
```diff
@@ -419,33 +421,43 @@ static void pxd_io_failover(struct work_struct *ws)
...
-	}
-
-	if (!pxd_dev->connected || pxd_dev->removing) {
-		printk_ratelimited(KERN_ERR"%s: pxd%llu: px is disconnected, failing IO.\n", __func__, pxd_dev->dev_id);
-		BIO_ENDIO(head->orig, -ENXIO);
-	} else {
-		// switch to native path
+		rc = pxd_initiate_failover(pxd_dev);
+		// If userspace cannot be informed of a failover event, force abort all IO.
+		if (rc) {
```
prabirpaul, line 447: This would abort IOs if rc==EBUSY i.e we have a failover op in progress. I don't think that's correct ?

dev.c
```diff
@@ -235,7 +235,13 @@ void fuse_request_send_nowait(struct fuse_conn *fc, struct fuse_req *req)
 	 */
 	rcu_read_lock();
 
-	if (fc->connected || fc->allow_disconnected) {
+	if (force) {
```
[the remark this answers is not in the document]
sulakshm: There is a slight difference. If force is set, request can get enqueued even if px is down and the wakeup needs posting only if px is connected.

pxd_fastpath.c
sulakshm: failover gets initiated only once by the first failing IO. Others just get enqueued to the failQ.

pxd_fastpath.c
sulakshm: pxd_request_{resume|suspend} is userspace suspend/resume op. Only once it can be called to set/clear state.
They are blocked for multiple calls within px-storage by device state. 
But internal to px-fuse suspend/resume can be called multiple times recursively. px-fuse does not call pxd_request_xxx calls.

prabirpaul: Ok, but pxd_request_suspend() uses a cas.

prabirpaul (approved):

pxd_fastpath.c
sulakshm: I see, will revise for keeping it similar.

sulakshm: I made both userspace request/suspend to be same. cas is not needed.

PR #154, opened 2020-09-22
export inprogress IO on a device

sulakshm:
Exporting in progress IO on a device for userspace px-storage to identify whether there are still any active IO on the device.
This is important while switching iopath between userspace and kernel on fastpath capable device.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1750,6 +1750,14 @@ static ssize_t pxd_debug_store(struct device *dev,
 	return count;
 }
 
+static ssize_t pxd_inprogress_show(struct device *dev,
+                     struct device_attribute *attr, char *buf)
+{
+	struct pxd_device *pxd_dev = dev_to_pxd_dev(dev);
+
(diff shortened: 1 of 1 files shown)


pxd.c
```diff
@@ -1770,6 +1779,7 @@ static struct attribute *pxd_attrs[] = {
 	&dev_attr_fastpath.attr,
 	&dev_attr_mode.attr,
 	&dev_attr_debug.attr,
+	&dev_attr_inprogress.attr,
```
prabirpaul, line 1782: Ok, but isn't this same as device "inflight" count?

prabirpaul (approved):

pxd.c
sulakshm: yes should be same, except for the window wherein IO is considered active. 
device inflight is controlled from within the px-fuse driver from places where it calls start_io_acct and end_io_acct.
The exported count is wholly private to the driver.

PR #156, opened 2020-09-30
px_fuse ioctl to abort ios

vjs-px:
new ioctl to abort all iOS
Note: this should be used as a last resort to avoid node restart when px is stuck

### DIFF
--- a/dev.c
+++ b/dev.c
@@ -1079,6 +1079,15 @@ __acquires(fc->lock)
 	end_requests(fc, &fc->processing);
 }
 
+void fuse_abort_all_ios(struct fuse_conn *fc)
+{
+	printk(KERN_INFO "PXD_IOCTL : Aborting all requests...");
+	spin_lock(&fc->lock);
+	fc->allow_disconnected = 0;
+	end_queued_requests(fc);
+	spin_unlock(&fc->lock);
+}
+
 static void fuse_conn_free_allocs(struct fuse_conn *fc)
 {
 	if (fc->per_cpu_ids)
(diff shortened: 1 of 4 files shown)


dev.c
```diff
@@ -1079,6 +1079,15 @@ __acquires(fc->lock)
 	end_requests(fc, &fc->processing);
 }
 
+void fuse_abort_all_ios(struct fuse_conn *fc)
+{
+	printk(KERN_INFO "PXD_IOCTL : Aborting all requests...");
```
prabirpaul, line 1084: Given the context and usage of this function, it should be in pxd.c alongside pxd_abort_context()?

prabirpaul (approved):

dev.c
vjs-px: end_queued_requests function is static and defined in dev.c, so I had to create another function in the same file to export it.

in v2.5.0, we have pxd_timeout(), which calls fuse_abort_conn() in dev.c, that in turn calls end_queued_requests()

I will leave this as is.

pxd.h
```diff
@@ -28,6 +28,7 @@
 #define PXD_IOC_DUMP_FC_INFO	_IO(PXD_IOCTL_MAGIC, 1)		/* 0x505801 */
 #define PXD_IOC_GET_VERSION		_IO(PXD_IOCTL_MAGIC, 2)		/* 0x505802 */
 #define PXD_IOC_INIT		_IO(PXD_IOCTL_MAGIC, 3)		/* 0x505803 */
+#define PXD_IOC_ABORT_IOS		_IO(PXD_IOCTL_MAGIC, 4)		/* 0x505804 */
```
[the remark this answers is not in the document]
vjs-px: ok, will change this

PR #159, opened 2020-10-05
fastpath fail scenario cleanup

sulakshm:
Fastpath volume operations failure cleanups.

If in the final stage of attaching a fastpath volume fails, then it cannot be immediately allowed to service IO in the native path, because SAN setup has put targets in fastpath mode, wherien the file handles are closed.
So they will remain in suspended state until a final cleanup request is issued.

Also do not ever update fastpath state without informed userspace.

This has been tested with userspace https://github.com/portworx/porx/pull/6505

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -181,6 +181,37 @@ static long pxd_ioctl_resize(struct file *file, void __user *argp)
 	return ret;
 }
 
+static long pxd_ioctl_fp_cleanup(struct file *file, void __user *argp)
+{
+	struct pxd_context *ctx = NULL;
+	struct pxd_fastpath_out cleanup_args;
+	long ret = 0;
+	struct pxd_device *pxd_dev;
+
+	if (copy_from_user(&cleanup_args, argp, sizeof(cleanup_args))) {
+		return -EFAULT;
+	}
+
+	if (cleanup_args.context_id >= pxd_num_contexts_exported) {
+		printk("%s : invalid context: %d\n", __func__, cleanup_args.context_id);
+		return -EFAULT;
+	}
+
+	ctx =  &pxd_contexts[cleanup_args.context_id];
+	if (!ctx || ctx->id >= pxd_num_contexts_exported) {
+		return -EFAULT;
+	}
+
+	pxd_dev = find_pxd_device(ctx, cleanup_args.dev_id);
+	if (pxd_dev != NULL) {
+		(void)get_device(&pxd_dev->dev);
+		ret = pxd_fastpath_vol_cleanup(pxd_dev);
+		put_device(&pxd_dev->dev);
+	}
+
+	return ret;
+}
+
 static long pxd_ioctl_run_user_queue(struct file *file)
 {
 	struct pxd_context *ctx = container_of(file->f_op, struct pxd_context, fops);
@@ -220,6 +251,8 @@ static long pxd_control_ioctl(struct file *file, unsigned int cmd, unsigned long
 		return pxd_ioctl_run_user_queue(file);
 	case PXD_IOC_RESIZE:
 		return pxd_ioctl_resize(file, (void __user *)arg);
+	case PXD_IOC_FPCLEANUP:
+		return pxd_ioctl_fp_cleanup(file, (void __user *)arg);
 	default:
 		return -ENOTTY;
 	}
--- a/pxd.c
+++ b/pxd.c
@@ -666,14 +666,14 @@ void pxd_process_ioswitch_complete(struct fuse_conn *fc, struct fuse_req *req,
 		spin_unlock(&pxd_dev->fp.fail_lock);
 	}
 
+	// reopen the suspended device
+	pxd_request_resume_internal(pxd_dev);
+
 	BUG_ON(atomic_read(&pxd_dev->fp.ioswitch_active) == 0);
 	atomic_set(&pxd_dev->fp.ioswitch_active, 0);
 
 	// reissue any failed IOs from local list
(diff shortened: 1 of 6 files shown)


pxd.c
```diff
@@ -181,6 +181,37 @@ static long pxd_ioctl_resize(struct file *file, void __user *argp)
 	return ret;
 }
 
+static long pxd_ioctl_fp_cleanup(struct file *file, void __user *argp)
```
sulakshm, line 184: a new ioctl to release volume in suspended state after attach failed in fastpath.

pxd.c
```diff
@@ -611,21 +644,6 @@ static int pxd_request(struct fuse_req *req, uint32_t size, uint64_t off,
 }
 #endif
 
-static inline unsigned int get_op_flags(struct bio *bio)
```
sulakshm, line 614: moved to header file.

pxd.c
```diff
@@ -1718,7 +1730,7 @@ static ssize_t pxd_debug_store(struct device *dev,
 		break;
 	case 'X': /* switch native path */
 		printk("dev:%llu - IO native path switch - ctrl failover\n", pxd_dev->dev_id);
-		pxd_switch_nativepath(pxd_dev);
+		pxd_debug_switch_nativepath(pxd_dev);
```
sulakshm, line 1733: renamed interface as debug

pxd_fastpath.c
```diff
@@ -416,6 +416,45 @@ static void pxd_failover_initiate(struct pxd_device *pxd_dev, struct pxd_io_trac
 	queue_work(pxd_dev->fp.wq, &head->wi);
 }
 
+static int remap_io_status(int status)
```
sulakshm, line 419: io status reconciliation from all replicas. Not all failures can be failedover... exceptions below, rest are converted to a hard -EIO error which triggers a failover, if allowed.

pxd_fastpath.c
```diff
@@ -1119,11 +1131,20 @@ void enableFastPath(struct pxd_device *pxd_dev, bool force)
 	memset(fp->device_path, 0, sizeof(fp->device_path));
 
 	pxd_dev->fp.fastpath = false;
-	pxd_resume_io(pxd_dev);
+	/// volume still remains suspended waiting for CLEANUP request to reopen IO.
```
sulakshm, line 1134: this is the case which needs a ioctl cleanup... because SAN path got established putting targets to handle IO in fastpath... but IO will immediately flow through native path without suspending and waiting for cleanup.

pxd_fastpath.c
```diff
@@ -1119,11 +1131,20 @@ void enableFastPath(struct pxd_device *pxd_dev, bool force)
...
 	printk(KERN_INFO"%s: Device %llu no backing volume setup, will take slow path\n",
 		__func__, pxd_dev->dev_id);
 }
 
+int pxd_fastpath_vol_cleanup(struct pxd_device *pxd_dev)
```
sulakshm, line 1139: this is the ioctl cleanup handler, a simple disable fastpath and resume io in native path.

pxd_fastpath.c
prabirpaul: So why can't pxd_fastpath_vol_cleanup() be called from here directly instead of relying on the ioctl?

pxd_fastpath.c
```diff
@@ -435,42 +474,45 @@ static void pxd_complete_io(struct bio* bio, int error)
...
+	if (!atomic_dec_and_test(&head->active)) {
+		// not all responses have come back
+		return;
+	}
+
+	// final reconciled status
+	blkrc = reconcile_io_status(head);
+
+	// debug condition for force fail
+	if (pxd_dev->fp.force_fail) blkrc = -EIO;
```
prabirpaul, line 514: Don't think the force_fail really needed? Can you not simulate this by other means for example, pulling a disk out from under the btrfs mountpoint?

pxd_fastpath.c
```diff
@@ -416,6 +416,45 @@ static void pxd_failover_initiate(struct pxd_device *pxd_dev, struct pxd_io_trac
...
+{
+	struct pxd_io_tracker *repl;
+	int status = 0;
+	int tmp;
+
+	BUG_ON(head->magic != PXD_IOT_MAGIC);
+	list_for_each_entry(repl, &head->replicas, item) {
+		BUG_ON(repl->magic != PXD_IOT_MAGIC);
+
+		tmp = remap_io_status(repl->status);
```
prabirpaul, line 444: A little confusing. If we are going to ignore anything != EIO, why call remap_io_status()?

sulakshm: That's not true. This function consolidates response status from each replica and translates response status to -EIO to indicate that a failover if possible should be initiated. All non zero status indicates a failure. A few of them is not to be considered a failure scenario for initiating failover, those are passed along directly to the end user without initiating failover.

pxd_fastpath.c
sulakshm: This is a debug interface to test, this can go away once all pieces are found stable.. Let it remain until independent testing completes. I will remove it then.

pxd_fastpath.c
sulakshm: If the px volume setup this late fails to setup in fastpath, then immediately setting the volume in native path, can start receiving IOs but the target has been configured for exporting target dev/file and file handle is closed. So processing this IO at the target will crash since IOs are received for closed file handles. px does a SAN cleanup and then reopens the px volume to allow IOs.

prabirpaul (approved):

PR #160, opened 2020-10-13
Add pxd tests

prabirpaul:
UTs for attach/detach, read/write operations.

### DIFF
--- a/test/pxd_test.cc
+++ b/test/pxd_test.cc
@@ -1,0 +1,419 @@
+#include <stdlib.h>
+#include <pxd.h>
+#include <gtest/gtest.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <linux/fuse.h>
+#include <sys/poll.h>
+#include <boost/iostreams/device/file_descriptor.hpp>
+#include <sys/uio.h>
+#include <string>
+#include <boost/lexical_cast.hpp>
+#include <functional>
+#include <linux/fs.h>
+#include <sys/ioctl.h>
+#include <thread>
+#include <vector>
+#include "fuse.h"
+
+using namespace std::placeholders;
+
+const char *pxd_op_names[] = {
+	"init",
+	"write",
+	"read",
+	"discard",
+	"flush",
+	"add,"
+	"pxd_remove",
+};
+
+std::string control_device(unsigned int driver_context_id)
+{
+	assert(driver_context_id < PXD_NUM_CONTEXTS);
+	std::string ret{PXD_CONTROL_DEV};
+	if (driver_context_id != 0)
+		ret += "-" + std::to_string(driver_context_id);
+	return ret;
+}
+
+const char *pxd_op_name(int opcode)
+{
+	if (opcode < PXD_INIT || opcode >= PXD_LAST) {
+		return "unknown";
+	} else {
+		return pxd_op_names[opcode - PXD_INIT];
+	}
+}
+
+class PxdTest : public ::testing::Test {
+protected:
+	int fd;		// control file descriptor
+	std::set<uint64_t> added_ids;
+	const size_t write_len = 4096;
+
+	PxdTest() : fd(-1) {}
+	virtual ~PxdTest() {
+		if (fd >= 0)
+			close(fd);
+	}
+
+	virtual void SetUp();
+	virtual void TearDown();
+
+	void dev_add(pxd_add_out &add, int &minor, std::string &name);
+	void dev_remove(uint64_t dev_id);
+	int wait_msg(int timeout); // timeout in seconds
+	void read_block(fuse_in_header *in, pxd_rdwr_in *rd);
+public:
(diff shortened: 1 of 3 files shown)


test/pxd_test.cc
```diff
@@ -0,0 +1,396 @@
...
+	ssize_t write_bytes = writev(fd, iov, 2);
+	ASSERT_GT(write_bytes, 0);
+
+	added_ids.insert(add.dev_id);
+
+	minor = write_bytes;
+	name = std::string(PXD_DEV_PATH) + std::to_string(add.dev_id);
+}
+
+int PxdTest::wait_msg(int timeout)
```
sulakshm, line 112: rename timeout with units please

test/pxd_test.cc
```diff
@@ -0,0 +1,396 @@
...
+	error = opcode;
+	len = sizeof(fuse_out_header) + op_len;
+}
+
+// Read block from kernel
+void PxdTest::read_block(fuse_in_header *hdr, pxd_rdwr_in *req)
+{
+	int iovcnt = 1;
+	struct iovec iov[iovcnt];
+	size_t iovlen = iovcnt * sizeof(iovec);
```
sulakshm, line 168: sizeof(iov) instead

test/pxd_test.cc
```diff
@@ -0,0 +1,396 @@
...
+
+	// Setup request header and payload
+	fuse_notify_header oh(PXD_READ_DATA, sizeof(pxd_read_data_out) + iovlen);
+	pxd_read_data_out rd_out = {hdr->unique, iovcnt, (uint32_t)req->offset};
+
+	memset(buf, 0, req->size);
+	iov[0].iov_base = buf;
+	iov[0].iov_len = req->size;
+	struct iovec wr_iov[3] = { { &oh, sizeof(oh) }, { &rd_out, sizeof(rd_out) },
+		{ iov, 1*sizeof(iovec) } };
```
sulakshm, line 180: iovlen instead of 1*sizeof(iovec)

test/pxd_test.cc
```diff
@@ -0,0 +1,396 @@
...
+
+	// Now read in the request from kernel
+	while (1) {
+		int ret = wait_msg(1);
+		ASSERT_EQ(0, ret);
+
+		read_bytes = read(fd, msg_buf, sizeof(msg_buf));
+		rdwr = reinterpret_cast<rdwr_in *>(msg_buf);
+
+		if (rdwr->in.opcode == PXD_READ) {
```
sulakshm, line 354: here many reads can come in... because of the reads from host machine on a new block device being made available. After device being added, first run the poller to throw out all IOs and complete it, then proceed with test IO and read it to replay it. That should be the reason for reads having different size than expected.

test/pxd_test.cc
```diff
@@ -0,0 +1,396 @@
...
+	assert(driver_context_id < PXD_NUM_CONTEXTS);
+	std::string ret{PXD_CONTROL_DEV};
+	if (driver_context_id != 0)
+		ret += "-" + std::to_string(driver_context_id);
+	return ret;
+}
+
+class PxdTest : public ::testing::Test {
+protected:
+	int fd;		// control file descriptor
```
sulakshm, line 32: would it be acceptable to rename this to ctl_fd.. test deals with too many file handles, so it helps to be more specific.

sulakshm (commented):
mostly minor comments.

test/pxd_test.cc
```diff
@@ -0,0 +1,396 @@
...
+	fd = open(control_device(0).c_str(), O_RDWR);
+	ASSERT_GT(fd, 0);
+
+	pxd_ioctl_init_args args;
+	auto ret = ioctl(fd, PXD_IOC_INIT, &args);
+	if (ret < 0) {
+		fprintf(stderr, "%s: init ioctl failed: %d(%s)", __func__, errno, strerror(errno));
+	}
+
+	auto read_bytes = static_cast<size_t>(ret);
```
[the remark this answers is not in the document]
prabirpaul: done

test/pxd_test.cc
prabirpaul: done

test/pxd_test.cc
prabirpaul: done

test/pxd_test.cc
prabirpaul: renamed

test/pxd_test.cc
prabirpaul: I'll address this in the next change. The issue is not the number of reads but the size. I see only one IO on dev as expected but the size is greater that what I asked for. Maybe some sysfs setting which is causing this, not sure yet.

test/pxd_test.cc
prabirpaul: done

PR #162, opened 2020-10-28
PWX-16752: Add kernel version check to use the right request op enums

jrivera-px:
Fix compilation issue 'REQ_OP_READ' undeclared  & REQ_OP_WRITE' undeclared 

/


sulakshm (approved):
Thanks, need to go into master and 2.7.0

PR #164, opened 2020-11-03
use blkdev_issue_discard for discard, dmthin does not support fallocate

maxkozlovsky:


io.c
```diff
@@ -1083,15 +1085,31 @@ static int io_discard(struct io_kiocb *req, const struct sqe_submit *s,
 
 	if (unlikely(!(req->file->f_mode & FMODE_WRITE))) {
 		ret = -EINVAL;
+		goto out;
+	}
+
+	inode = req->file->f_inode;
+	if (S_ISBLK(inode->i_mode)) {
+		ret = blkdev_issue_discard(inode->i_bdev, off / SECTOR_SIZE,
```
prabirpaul, line 1093: I don't think this is going to work. If there's a dirty page in cache for this offset, the discard will get queued ahead of the write.

prabirpaul (approved):

PR #167, opened 2020-11-23
PWX-17237: px-fastpath extend node wipe

sulakshm:
if volumes are attached in fastpath, there is a kernel file handle opened on that target.
This fails node wipe. Have extended a sysfs 'release' object on px volume, which when written with a magic keyword will
disable fastpath on the volume.

Goes along with: https://github.com/portworx/porx/pull/6942

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1807,7 +1807,7 @@ static int pxd_nodewipe_cleanup(struct pxd_context *ctx)
 static ssize_t pxd_release_store(struct device *dev,
 			struct device_attribute *attr, const char *buf, size_t count)
 {
-	static const char wipemagic[] = "PWXWIPE";
+	static const char wipemagic[] = "P0RXR3l3@53";
 	int i;
 	struct pxd_context *ctx;
 
--- a/pxd.c
+++ b/pxd.c
@@ -57,6 +57,7 @@ module_param(pxd_num_contexts_exported, uint, 0644);
 module_param(pxd_num_contexts, uint, 0644);
 module_param(pxd_detect_zero_writes, uint, 0644);
 
+static void pxd_abort_context(struct work_struct *work);
 static int pxd_bus_add_dev(struct pxd_device *pxd_dev);
 
 struct pxd_context* find_context(unsigned ctx)
@@ -1801,6 +1802,10 @@ static ssize_t pxd_inprogress_show(struct device *dev,
 static void pxd_nodewipe_cleanup(struct pxd_context *ctx)
 {
 	struct list_head *cur, *tmp;
+
+	cancel_delayed_work_sync(&ctx->abort_work);
+	pxd_abort_context(&ctx->abort_work.work);
+
 	spin_lock(&ctx->lock);
 	list_for_each_safe(cur, tmp, &ctx->list) {
 		struct pxd_device *pxd_dev = container_of(cur, struct pxd_device, node);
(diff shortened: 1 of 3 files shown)


pxd.c
```diff
@@ -1776,6 +1778,58 @@ static ssize_t pxd_inprogress_show(struct device *dev,
...
+	}
+	spin_unlock(&ctx->lock);
+
+	return 0;
+}
+
+static ssize_t pxd_release_store(struct device *dev,
+			struct device_attribute *attr, const char *buf, size_t count)
+{
+	static const char wipemagic[] = "P0RXR3l3@53";
```
prabirpaul, line 1810: - Wouldn't it be more appropriate to extend the pxd_iotcl() to do the job?

More importantly can this be not done as a part of module unload? Same issue exists for stuck pxd IOs. The way nodewipe deals with them is by simply setting the timeout to 30secs. If you just add disableFastPath() to module cleanup, that should do it. Also this way px-storage UTs cleanup wouldn't need to do anything special.

sulakshm: ioctl needs to open pxd_control and this is being interpreted as a userspace attach to px-fuse. So ioctl cannot be used from node wipe. If IOs are stuck to virtual pxd devices, they still do not hold direct references to backing devices, node wipe can proceed to pull the rug under them and will pass. But with fastpath active, there is a direct reference held that needs to be broken before node wipe can pass. They are not similar.

prabirpaul: No, ioctl has a reserved channel.
What do u think about the other option - doing it in module unload?

sulakshm: Aborting IOs on abort timeout doesn't switch IO path... io path switch can happen only with  px-storage  being aware otherwise other targets will be unaware....  And while fastpath is active, IO being failed happens without any intervention..., this timeout run only ensures newer IOs do fail immediately.
Infact node wipe already uses timeoutDir to reset/force abort handler to run... extending it to release fastpath is more inline to current design.
Ioctl may work,  but need to aware of the proper device to open which may fail if module param pxd_num_contexts_exported gets used.

Can we not use the current posted method as is... i see no benefits vs other approaches to current approach.

pxd.c
```diff
@@ -1776,6 +1778,55 @@ static ssize_t pxd_inprogress_show(struct device *dev,
 	return sprintf(buf, "%d", atomic_read(&pxd_dev->ncount));
 }
 
+static int pxd_nodewipe_cleanup(struct pxd_context *ctx)
```
sulakshm, line 1781: revised method to not force abort context...

pxd.c
```diff
@@ -1776,6 +1778,55 @@ static ssize_t pxd_inprogress_show(struct device *dev,
...
+	}
+	spin_unlock(&ctx->lock);
+
+	return 0;
+}
+
+static ssize_t pxd_release_store(struct device *dev,
+			struct device_attribute *attr, const char *buf, size_t count)
+{
+	static const char wipemagic[] = "P0RXR3l3@53";
```
prabirpaul, line 1807: Writing a magic code to a control device is basically what an ioctl is. We already have that mechanism, with 10 ioctls implemented. If this interface is insufficient, then it should be extended. If it is incorrect, then a new one should be implemented and everyone migrated.
It's up to you but I don't agree with this roundabout way.

prabirpaul (approved):

PR #174, opened 2020-12-15
PWX-16992: Fix compile issues for 5.9.6-1.el7.elrepo.x86_64 kernel.

jrivera-px:
Compilation fixes for changes 5.9.x kernel changes.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -41,7 +41,11 @@
 #define PXD_TIMER_SECS_DEFAULT 600
 #define PXD_TIMER_SECS_MAX (U32_MAX)
 
+#if defined(PF_LESS_THROTTLE)
 #define PF_IO_FLUSHER (PF_MEMALLOC_NOIO | PF_LESS_THROTTLE)
+#elif defined(PF_LOCAL_THROTTLE)
+#define PF_IO_FLUSHER (PF_MEMALLOC_NOIO | PF_LOCAL_THROTTLE)
+#endif
 #define IS_SET_IO_FLUSHER_FLAG(flg) ((flg & PF_IO_FLUSHER) == PF_IO_FLUSHER)
 #define IS_SET_IO_FLUSHER(task) (IS_SET_IO_FLUSHER_FLAG((task->flags)))
 #define SET_IO_FLUSHER(task) (task->flags |= PF_IO_FLUSHER)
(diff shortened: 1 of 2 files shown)


pxd.c
```diff
@@ -582,8 +586,11 @@ int pxd_handle_device_limits(struct fuse_req *req, uint32_t *size, uint64_t *off
 		}
 
 		bio_chain(b, req->bio);
+#if LINUX_VERSION_CODE <= KERNEL_VERSION(5,8,0)
 		generic_make_request(b);
-
+#else		
+		submit_bio_noacct(b);
+#endif
```
sulakshm, line 593: this is correct change.

pxd.c
```diff
@@ -261,8 +265,8 @@ static void print_io_flusher_state(unsigned int new_flags,
 
 static int is_io_flusher_supported(void)
 {
-	#if LINUX_VERSION_CODE >= KERNEL_VERSION(3,10,0) && LINUX_VERSION_CODE <= KERNEL_VERSION(5,8,0)
-		#if defined(PF_MEMALLOC_NOIO) && defined(PF_LESS_THROTTLE)
+  	#if LINUX_VERSION_CODE >= KERNEL_VERSION(3,10,0) && defined(PF_MEMALLOC_NOIO)
+                #if defined(PF_LESS_THROTTLE) || defined(PF_LOCAL_THROTTLE)
```
sulakshm, line 269: the above macro seems incorrect, checking same entity between OR?

jrivera-px: Checking the same entity?   I am sorry I don't understand what you mean?

pxd_fastpath.c
```diff
@@ -1470,7 +1470,11 @@ void pxd_fastpath_adjust_limits(struct pxd_device *pxd_dev, struct request_queue
 			if (bque) {
 				printk(KERN_INFO"pxd device %llu queue limits adjusted with block dev %p(%s)\n",
 					pxd_dev->dev_id, bdev, bdevname(bdev, name));
+#if LINUX_VERSION_CODE <= KERNEL_VERSION(5,8,0)
 				blk_queue_stack_limits(topque, bque);
+#else
+				blk_stack_limits(&topque->limits, &bque->limits, 0);
```
sulakshm, line 1476: this is fine too.

jrivera-px: Thanks...

sulakshm (approved):
just one minor nit to improve readability. otherwise looks good.

pxd.c
vjs-px: PF_LESS_THROTTLE is a generic kernel #define which is defined include/linux/sched.h
 PF_LESS_THROTTLE and PF_LOCAL_THROTTLE have different meaning and implementation, I don't think we should be assigning PF_LESS_THROTTLE = PF_LOCAL_THROTTLE

vjs-px: We can change the implementation such that all the new code related to IO_FLUSHER is compile time enabled based on 
#ifdef PF_IO_FLUSHER
#endif

This can be taken up as a new PR

vjs-px (approved):

PR #177, opened 2021-01-07
PWX-18027: fix fuse 'req' leak

sulakshm:

# issue #
when fastpath devices are switched iopath, the fuse context that gets used for communication between porx and px-fuse does not get freed resulting in following exception during rmmod eventually.

```
[Thu Jan  7 16:34:30 2021] BUG pxd_fuse_request (Tainted: G           OE    ): Objects remaining in pxd_fuse_request on __kmem_cache_shutdown()
```

### DIFF
--- a/dev.c
+++ b/dev.c
@@ -210,18 +210,16 @@ static void request_end(struct fuse_conn *fc, struct fuse_req *req,
 	int status)
 {
 	u64 uid = req->in.unique;
-#ifdef __PX_BLKMQ__
-	bool using_blkque = req->pxd_dev->using_blkque;
-#endif
+	bool shouldfree = false;
+
+	/* 'req->end' is always set if the context gets used.
+	 * if error happens during processing, error path handling does free request.
+	 */
 	if (req->end)
-		req->end(fc, req, status);
+		shouldfree = req->end(fc, req, status);
 	fuse_put_unique(fc, uid);
 
-#ifndef __PX_BLKMQ__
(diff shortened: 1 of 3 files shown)


dev.c
```diff
@@ -210,18 +210,16 @@ static void request_end(struct fuse_conn *fc, struct fuse_req *req,
...
-		req->end(fc, req, status);
+		shouldfree = req->end(fc, req, status);
 	fuse_put_unique(fc, uid);
 
-#ifndef __PX_BLKMQ__
-	fuse_request_free(req);
-#else
-	if (!using_blkque) fuse_request_free(req);
-#endif
+	if (shouldfree) fuse_request_free(req);
```
prabirpaul, line 222: Instead of this, can we not do something like a safe_free(req)?

sulakshm: All caller points are clearly defined and extended function about freeing fuse_request is unambiguous.
Whats the concern? And what is safe_free?

prabirpaul: I meant this:
```
	if (req->end)
		req->end(fc, req, status);
	fuse_put_unique(fc, uid);
        if (req)
                fuse_request_free(req);
...

#ifndef __PX_BLKMQ__
	blk_end_request(req->rq, status, blk_rq_bytes(req->rq));
#else
	blk_mq_end_request(req->rq, errno_to_blk_status(status));
	req = NULL;
#endif
```

prabirpaul (approved):

PR #202, opened 2021-03-30
fix master px-fuse broken build

sulakshm:

Fix master px-fuse submodule for compilation.
Compile tested master. 
jenkins run failed for some other issue.
https://jenkins.portworx.dev/job/DEV/job/Porx-02/642/console

Brings in the following changes from private branch to master:
```
commit 47961a204d6ae1847bb5490467e4d06154071731 (HEAD -> origin/iorthread1, origin/iorthread1)
Author: Maxim Kozlovsky <mkozlovsky@purestorage.com>
Date:   Tue Jan 5 09:23:41 2021 -0800

    ioring offload thread support

commit c6ee65f50dd0458514571fc65637abe68c61dbd1
Author: Maxim Kozlovsky <mkozlovsky@purestorage.com>
Date:   Fri Dec 4 11:48:17 2020 -0800

    support for passing parameters to initialize ioring context

```

### DIFF
--- a/io.c
+++ b/io.c
@@ -1954,6 +1954,8 @@ static int io_sq_thread(void *data)
 	set_fs(USER_DS);
 #endif
 
+	pr_info("%s: started to %d", __func__, ctx->sq_thread_idle);
+
 	timeout = inflight = 0;
 	while (!kthread_should_park()) {
 		bool all_fixed, mm_fault = false;
--- a/io.c
+++ b/io.c
@@ -236,20 +236,53 @@ static void io_ring_ctx_ref_free(struct percpu_ref *ref)
 	complete(&ctx->ctx_done);
 }
 
-static int io_ring_ctx_init(struct io_ring_ctx *ctx)
+static size_t queue_size(struct io_ring_ctx *ctx)
+{
+	return sizeof(struct fuse_queue_cb) + sizeof(struct fuse_queue_cb) +
+			ctx->sq_entries * sizeof(struct io_uring_sqe) +
+			ctx->cq_entries * sizeof(struct io_uring_cqe);
+}
+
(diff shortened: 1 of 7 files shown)


io.h
```diff
@@ -75,7 +75,12 @@ struct io_ring_ctx {
 		unsigned		sq_mask;
 		unsigned		sq_thread_idle;
 
-		struct io_ring_queue	*queue;
+		void    *queue;
```
prabirpaul, line 78: What happened here? Did io_ring_queue get deprecated on certain kernels? 
All changes look fine, a note regarding why this is required would be great.

io.c
```diff
@@ -2685,17 +2699,46 @@ static int io_uring_release(struct inode *inode, struct file *file)
...
+		wake_up(&ctx->sqo_wait);
+		return 0;
 	case PXD_IOC_RUN_IO_QUEUE:
 		return io_run_queue(ctx);
 	case PXD_IOC_REGISTER_FILE:
 		return io_sqe_register_file(ctx, arg);
 	case PXD_IOC_UNREGISTER_FILE:
 		return io_sqe_unregister_file(ctx, arg);
+	case PXD_IOC_INIT_IO:
+		return io_ring_ioctl_init(ctx, arg);
```
prabirpaul, line 2741: Why the new ioctls, and why the move out of io_uring_open() ?

sulakshm: these are the missed changes.
io uring now has additional params through which the queue depth etc can be setup.
So there will be an initial ioctl INIT_IO to set this up.
And then usage will be as usual. Earlier it was single step with default hardcoded params.

io.h
sulakshm: This is px-fuse defined struct. This has been reorg for the same reason to split into config init and uring use.

prabirpaul (approved):

PR #213, opened 2021-07-01
PWX-20602: pre-check stale devices before add

sulakshm:

Adds stale device checks before adding new device to kernel.
checks for both actual block device and its sysfs device existence.
if they are present, during a device attach, then device attach will crash in add disk path inside the kernel.
This window may open up if forced detach gets used.

Closes #
part of https://portworx.atlassian.net/browse/PWX-20602

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1354,14 +1354,16 @@ ssize_t pxd_add(struct fuse_conn *fc, struct pxd_add_ext_out *add)
 	err = vfs_stat(devfile, &pxdev_stat);
 	if (err == 0) {
 		pr_err("stale device(%s) found, attach fail", devfile);
-		return -EEXIST;
+		err = -EEXIST;
+		goto out_module;
 	}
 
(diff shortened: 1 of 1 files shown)


pxd.c
```diff
@@ -1347,6 +1349,21 @@ ssize_t pxd_add(struct fuse_conn *fc, struct pxd_add_ext_out *add)
 		return pxd_dev->minor;
 	}
 
+	/* pre-check to detect if prior instance is removed */
+	sprintf(devfile, "/dev/pxd/pxd!pxd%llu", add->dev_id);
+	err = vfs_stat(devfile, &pxdev_stat);
+	if (err == 0) {
+		pr_err("stale device(%s) found, attach fail", devfile);
+		return -EEXIST;
```
maxkozlovsky, line 1357: goto appropriate cleanup label instead of return

sulakshm: revised.

maxkozlovsky (approved):

PR #215, opened 2021-07-02
io profiling: high level io density map support

sulakshm:
a simple io profiler per volume at driver to identify frequent request sizes.
helps plot out a density map of requests on size for read/write and discard.

Closes #

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -598,6 +598,24 @@ static void pxd_req_misc(struct fuse_req *req, uint32_t size, uint64_t off,
 					  ((flags & REQ_FUA) ? PXD_FLAGS_FUA : 0) |
 					  ((flags & REQ_META) ? PXD_FLAGS_META : 0);
 #endif
+
+	if (size != 0) {
+		// Add logs about the requests to determine IOs are generally of
+		// what particular block sizes
+		unsigned idx = (size / PXD_LBS) - 1;
+		if (idx >= 256) idx = 255;
+		switch (req->in.opcode) {
+		case PXD_READ:
+			req->pxd_dev->io_map[0][idx]++;
+		case PXD_WRITE:
+			req->pxd_dev->io_map[1][idx]++;
+		case PXD_DISCARD:
+			req->pxd_dev->io_map[2][idx]++;
+			break;
+		default:
+			break;
+		}
+	}
 }
 
 /*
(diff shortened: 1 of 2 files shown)


maxkozlovsky (commented):
I think according to Linux design things like this are expected to be done through the dtrace style processing of kernel tracepoints. The block device traces provide enough information that the histograms of i/o sizes can be collected entirely from the user space without adding any overhead to the normal request processing on as needed basis. There are probably tools to do that already.

sulakshm:
blktrace exists but it generates voluminous amount of very minute details from the blk layer. We may not require such indepth information. And in any case most of the time spent processing request is with the userspace, we already know it, after the IO is dispatched. This is to build information about the io sizes, which may help optimize our configuration easier and also understand app IO patterns wrt data size.

maxkozlovsky:
Collecting the traces with blktrace is not necessary, the information can be assembled by using the tracepoints with bpf programs. Here is the output of bitesize-bpfcc which comes with bpfcc-tools. It does exactly what you are doing and more only without custom kernel code and without adding extra overhead when the data collection is not required.

Tracing block I/O... Hit Ctrl-C to end.
^C
Process Name = systemd-udevd
     Kbytes              : count     distribution
         0 -> 1          : 0        |                                        |
         2 -> 3          : 0        |                                        |
         4 -> 7          : 32       |****************************************|
         8 -> 15         : 4        |*****                                   |
        16 -> 31         : 16       |********************                    |
        32 -> 63         : 13       |****************                        |
        64 -> 127        : 7        |********                                |
       128 -> 255        : 3        |***                                     |

Process Name = umount
     Kbytes              : count     distribution
         0 -> 1          : 0        |                                        |
         2 -> 3          : 0        |                                        |
         4 -> 7          : 0        |                                        |
         8 -> 15         : 0        |                                        |
        16 -> 31         : 4        |****************************************|
        32 -> 63         : 4        |****************************************|
        64 -> 127        : 2        |********************                    |

Process Name = kworker/u16:5
     Kbytes              : count     distribution
         0 -> 1          : 7        |*************************               |
         2 -> 3          : 0        |                                        |
         4 -> 7          : 11       |****************************************|
         8 -> 15         : 1        |***                                     |
        16 -> 31         : 0        |                                        |
        32 -> 63         : 2        |*******                                 |

sulakshm:
Ok, this changeset is not all that important. I shall take this back.

PR #216, opened 2021-07-06
Ln/fuseopt

sulakshm:
Second part of master branch performance improvement.
This patch focusses on improving IO perf between px driver and the userspace, both ways.

Performance has improved manifold across the board.

Closes #

image

### DIFF
--- a/dev.c
+++ b/dev.c
@@ -1251,25 +1251,27 @@ static int fuse_process_user_queue(void *c)
 	unsigned long spin_wait = jiffies;
 
 	while (!kthread_should_stop()) {
-		if (signal_pending(current))
-			flush_signals(current);
-
 		do {
 			if (user_request_pending(fc)) break;
 			cpu_relax();
 		} while (jiffies < spin_wait);
 
 		// prepare to sleep
-		if (!user_request_pending(fc) && cur_mm != NULL) {
-			fuse_remove_user_access(cur_mm, old_fs);
-			cur_mm = NULL;
-		}
+		if (!user_request_pending(fc)) {
+			if (cur_mm != NULL) {
+				fuse_remove_user_access(cur_mm, old_fs);
+				cur_mm = NULL;
+			}
 
-		atomic_inc(&cb->r.need_wake_up);
-		wait_event_interruptible(fc->io_wait,
+			atomic_inc(&cb->r.need_wake_up);
+			if (signal_pending(current))
+				flush_signals(current);
+
+			wait_event_interruptible(fc->io_wait,
 				(user_request_pending(fc) || kthread_should_stop() ||
 				 kthread_should_park()));
-		atomic_dec(&cb->r.need_wake_up);
+			atomic_dec(&cb->r.need_wake_up);
+		}
 
 		if (kthread_should_stop()) {
 			break;
@@ -1304,9 +1306,12 @@ void fuse_pause_user_queue(struct fuse_conn *fc)
 
 	pr_info("parking worker threads");
 	for (i=0; i<NWORKERS; i++) kthread_park(fc->io_worker_thread[i]);
+	BUG_ON(atomic_read(&cb->r.need_wake_up) != 0);
 
-	if (fc->user_mm) mmdrop(fc->user_mm);
-	BUG_ON(atomic_read(&cb->r.need_wake_up) == 0);
+	if (fc->user_mm) {
+		mmdrop(fc->user_mm);
+		fc->user_mm = NULL;
+	}
 }
 
 void fuse_restart_user_queue(struct fuse_conn *fc)
@@ -1330,16 +1335,6 @@ int fuse_conn_init(struct fuse_conn *fc)
 	init_waitqueue_head(&fc->waitq);
 	fc->request_map = kmalloc(FUSE_MAX_REQUEST_IDS * sizeof(struct fuse_req*),
 		GFP_KERNEL);
-
-	init_waitqueue_head(&fc->io_wait);
-	spin_lock_init(&fc->io_lock);
-	for (i=0; i<NWORKERS; i++) {
-		fc->io_worker_thread[i] = kthread_create(fuse_process_user_queue, fc, "userq-worker-%d", i);
-		if (IS_ERR(fc->io_worker_thread[i])) {
-			rc = (int) PTR_ERR(fc->io_worker_thread[i]);
-			goto err_out;
-		}
-	}
 
 	rc = -ENOMEM;
 	if (!fc->request_map) {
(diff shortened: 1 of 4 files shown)


maxkozlovsky (commented):
Using extra threads for this is an overkill. While it may increase the performance it does not justify the extra cpu cost for spinning kernel threads (which basically removes the core from processing any other work). 

The amount of the work done here is pretty minimal, it is copy data and call bio_done() for the read request and call bio_done() for write request. 8 threads is way too much overhead for the work performed.

What happens if the response is done directly in the original user thread context without going through the response queue as in the original code?

We should get rid of thus driver specific response queue anyway while we can to avoid backward compatibility issues. Any work done there should be merged with the iouring queue. The read response eventually would be responded from the network code directly for the remote reads and from the io completion for local requests.

sulakshm:
1/ We are already looking at perf regression in IO path and it does not scale. This changeset tries to remove the bottlenecks.
The added resources are active only when IOs are to be serviced.

2/ The amount of work may be minimal, but the number of requests will scale as more devices get attached. 
Also, px need not be involved in this phase of completion. This change takes complete ownership of completing requests away from the px userspace process.

3/ I had already tried that approach. It still does not scale as expected, with 2.8 as baseline and ctrl reply as target, see below.
Taking randwrites out, "all ctrl reply" is 3 above par, 3 below par profiles,
while "posted patch" is 5 above par, 1 below (1 vol seq write).

```
KIOPS 

nvol=1     | 2.8 | master | all ctrl reply | perf patch
write        | 172 | 115 | 125 | 113  (-par)
read         | 246 | 246 | 328 | 325  (+par)
randwrite | 87 | 68 | 79 | 67         (-par - marginal)
randread  | 137 | 93 | 85 | 278    (--par)

nvol=40    | 2.8 | master | master all ctrl reply | perf patch
write          | 234 | 312 | 325 | 422 (+par)
read          | 420 | 184 | 314 | 515 (--par)
randwrite  | 95 | 81 | 70 | 82   (-par marginal)
randread   | 62 | 73 | 111 | 506 (+par)

```


4/ Any merging of response path with a iouring is beyond the scope of this patch. Also my understanding is read responses do carry more information than data (like checksum and timestamps). So it has to be conditional and more logic from userspace needs to get exposed to get this path ready. For writes again, there may be replicated volumes whose response/quorum etc logic needs to come in driver code before response. This is not the focus of this PR.

maxkozlovsky:
>>  I had already tried that approach. It still does not scale as expected

Could you please explain why the same code scales for 2.8 and does not scale for the master? What is the difference?

sulakshm:
> > > I had already tried that approach. It still does not scale as expected
> 
> Could you please explain why the same code scales for 2.8 and does not scale for the master? What is the difference?

Two significant reasons why 2.8 does better in _some_ profiles.
1/ there is a higher percentage of merges happening at the physical drive end. This more than compensates for the slow path in driver and target side IO paths.

2/ master as is, tries to service all requests in a single thread from userspace in a tight loop (run_queue).
This does not scale. The same is the reason with the driver end as well.

maxkozlovsky:
The question is why taking the same code path as 2.8 to reply to requests does not scale in master?

sulakshm:
How is it the same code path?

In 2.8, when one looks at the complete IO path (still simplified)
a) ctrl device fetch, 
b) io scheduling at coordinator
c) reach target through messenger
d) kaio module for IO at target
e) ctrl device response to complete IO.

vs
in master
a) memory mapped IO fetch
b) io scheduling at coordinator.
c) reach target through messenger
d) uring module for IO at target
e) memory mapped IO completion.


Now completing IO like 2.8 from master, still does not match 2.8.
Additionally, master has direct IO for random reads.

Also, in 2.8 new requests are fetched only when the threadpool has no new requests.
In master, new requests are fetched all the time.

dev.c
```diff
@@ -1135,7 +1158,175 @@ static void fuse_conn_queues_init(struct fuse_conn_queues *queue)
...
+		if (unlikely(mm_fault)) {
+			fuse_user_complete(fc, ureq[i].unique, ureq[i].user_data, -EFAULT);
+		} else {
+			fuse_process_user_request(fc, &ureq[i]);
+		}
+	}
+
+}
+
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 11, 0)
```
sulakshm, line 1197: wrappers to control userspace locking.

dev.c
```diff
@@ -1135,7 +1158,175 @@ static void fuse_conn_queues_init(struct fuse_conn_queues *queue)
...
+	int i;
+
+	mmgrab(current->mm);
+	fc->user_mm = current->mm;
+	pr_info("unparking worker threads");
+	for (i=0; i<NWORKERS; i++)
+		if (fc->io_worker_thread[i]) kthread_unpark(fc->io_worker_thread[i]);
+}
+
+int fuse_conn_init(struct fuse_conn *fc, uint32_t max_workers)
```
sulakshm, line 1329: 'max_workers's argument gets passed part of initialization. This is driven through a module param. Default is zero.

pxd.c
```diff
@@ -64,15 +64,26 @@ uint32_t pxd_num_contexts = PXD_NUM_CONTEXTS;
 uint32_t pxd_num_contexts_exported = PXD_NUM_CONTEXT_EXPORTED;
 uint32_t pxd_timeout_secs = PXD_TIMER_SECS_DEFAULT;
 uint32_t pxd_detect_zero_writes = 0;
+uint32_t pxd_offload = 0;
 
 module_param(pxd_num_contexts_exported, uint, 0644);
 module_param(pxd_num_contexts, uint, 0644);
 module_param(pxd_detect_zero_writes, uint, 0644);
+/// specify number of threads for bgio processing
+module_param(pxd_offload, uint, 0644);
```
sulakshm, line 73: the above controls the number of kernel threads for offloading req processing in the kernel. By default disabled.

pxd.h
```diff
@@ -214,14 +214,17 @@ struct pxd_device* find_pxd_device(struct pxd_context *ctx, uint64_t dev_id);
...
 
+uint32_t pxd_offload_threads(void);
 static inline
 int pxd_supported_features(void)
 {
 	int features = 0;
 #ifdef __PX_FASTPATH__
 	features |= PXD_FEATURE_FASTPATH;
 #endif
+	if (pxd_offload_threads()) features |= PXD_FEATURE_BGIO;
```
sulakshm, line 227: kernel offloading support is exported to userspace as a feature flag.

sulakshm (commented):
added comment.

PR #217, opened 2021-07-06
PWX20756: iouring increase file limit

sulakshm:

https://portworx.atlassian.net/browse/PWX-20756

hit register file exception as below
```
#0  0x00007f387a1a418b in raise () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f387a183859 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f387a183729 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007f387a194f36 in __assert_fail () from /lib/x86_64-linux-gnu/libc.so.6
#4  0x000056005d0a63ee in px::ioring::Context::register_file (
    this=<optimized out>, fd=<optimized out>) at gdsd/pxd_io.cc:144
#5  0x000056005d0e2c79 in px::target::FileHandle::FileHandle (
```

The number of replica targets is 521 on that node.
Px in master, opens each file in two modes this doubles the count.
In px-fuse driver for iouring, there is a max limit of only 1024 file handles. This breaches the limit failing the px startup code.
The px-fuse limit for file descr limit has to be increased.

Closes # https://portworx.atlassian.net/browse/PWX-20756

### DIFF
--- a/io.c
+++ b/io.c
@@ -2131,19 +2131,15 @@ static int io_sqe_register_file(struct io_ring_ctx *ctx, int fd)
 	}
 
 	err = -ENFILE;
-	if (i == ctx->nr_user_files) {
-		pr_err("iouring: max user file limit");
+	if (i == ctx->nr_user_files)
 		goto out;
-	}
 
 	err = -EBADF;
 	ctx->user_files[i] = fget(fd);
 	if (!ctx->user_files[i])
-		pr_err("iouring: bad file descriptor");
 		goto out;
 
 	if (ctx->user_files[i]->f_op == &fuse_dev_operations) {
-		pr_err("iouring: got control fd");
 		fput(ctx->user_files[i]);
 		ctx->user_files[i] = NULL;
 		goto out;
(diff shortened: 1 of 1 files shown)


io.c
```diff
@@ -2131,15 +2131,19 @@ static int io_sqe_register_file(struct io_ring_ctx *ctx, int fd)
...
-	if (i == ctx->nr_user_files)
+	if (i == ctx->nr_user_files) {
+		pr_err("iouring: max user file limit");
 		goto out;
+	}
 
 	err = -EBADF;
 	ctx->user_files[i] = fget(fd);
 	if (!ctx->user_files[i])
+		pr_err("iouring: bad file descriptor");
```
maxkozlovsky, line 2142: Missing {} will always goto out.

maxkozlovsky (commented):
You could add printf for the error code into the user space code calling register file instead of multiple printfs here.

io.c
sulakshm: yes revised.

maxkozlovsky (approved):

PR #225, opened 2021-08-16
PWX-20267: fix compilation from master c220a5a8 for 5.11.x kernels

jrivera-px:
v2.7 & v2.8 do fuse does not compile under the 5.11 kernel.   We need this to support fc33 latest kernel releases on OKD for PX 2.9.x.

PWX-20576

This is pulled from @sulakshm change: https://github.com/portworx/px-fuse/commit/c220a5a8faa192f5b9cf581911870ac55e7efe87.  There was another change associated with 5.11 in that PR which was made in file io.c but file does not exist in this branch so I left it out.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -298,11 +298,25 @@ static const struct block_device_operations pxd_bd_ops = {
 	.release		= pxd_release,
 };
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,11,0)
+static void pxd_update_stats(struct fuse_req *req, int sgrp, unsigned int count)
+#else
 static void pxd_update_stats(struct fuse_req *req, int rw, unsigned int count)
+#endif
 {
 		struct pxd_device *pxd_dev = req->queue->queuedata;
+		
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,11,0) || defined(__EL8__)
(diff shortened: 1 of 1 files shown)


prabirpaul (approved):
Would be good to get a bvt run on a >= 5.11 kernel. That would  test out both stats and the resize path.

jrivera-px:
This PR only exists in v2.10.0 portworx/px-fuse which is submodules into "master".   It does not exist in  v2.8.0 or v2.9.0 of portworx/px-fuse.   Currently gs-rel-2.8/2.9 px-fuse submodule is point v2.7.0.

sulakshm:
@jrivera-px just so you know gs-rel-2.10.0-ea porx branch uses px-fuse with label v2.10.0

jrivera-px:
@sulakshm,  might have misunderstood your comment but I don't see these changes in the v2.7.0 branch of px-fuse?  The last commit I see on pxfuse v2.7.0 is a96c121ce5fc40b34cbd8ee249928efde4cbafb7 which is what the submodule is set to in PX gs-rel-2.7.0, 2.8.0 & 2.9.0?  So I believe kernels 5.11 and above will still fail to build.  I definitely see the change in v2.10.0 pxfuse.

PR #226, opened 2021-08-19
add ioctl for registering fixed buffers with iouring

maxkozlovsky:
add support for providing partial region and extending the mapping
later while reusing the same fixed region.

Closes #

### DIFF
--- a/io.c
+++ b/io.c
@@ -2381,8 +2381,10 @@ static int io_sqe_register_buffers(struct io_ring_ctx *ctx, void __user *uarg)
 			for (j = 0; j < pret; j++)
 				put_page(pages[j]);
 		}
-		kvfree(imu->bvec);
-		imu->bvec = NULL;
+		if (!existing) {
+			kvfree(imu->bvec);
+			imu->bvec = NULL;
+		}
 		goto err;
 	}
 
--- a/io.c
+++ b/io.c
@@ -195,13 +195,6 @@ struct io_cq_ring {
 	struct io_uring_cqe	cqes[];
 };
 
-struct io_mapped_ubuf {
-	u64		ubuf;
-	size_t		len;
-	struct		bio_vec *bvec;
-	unsigned int	nr_bvecs;
-};
-
 #define IO_PLUG_THRESHOLD		2
 #define IO_IOPOLL_BATCH			8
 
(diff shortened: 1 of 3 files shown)


io.c
```diff
@@ -2244,6 +2237,166 @@ static int io_sqe_files_register(struct io_ring_ctx *ctx, void __user *arg,
...
+		goto err;
+	}
+
+	off = ubuf & ~PAGE_MASK;
+	size = arg.len;
+	for (j = 0; j < nr_pages; j++) {
+		size_t vec_len;
+
+		vec_len = min_t(size_t, size, PAGE_SIZE - off);
+		imu->bvec[j + offset].bv_page = pages[j];
```
sulakshm, line 2377: if this is a previously mapped range, then previous mapping is overwritten here for the new range passed. This would leave previous page pinned in memory and references lost to it.

Also uring allows only initializing one set of fixed buffers. This currently extends to allow 16 such sets.

sulakshm (changes requested):
What's the focus here?
1/ is the plan to support multiple fixed buffers ? This is okay but is a deviation from uring which allows only one set of buffers to be registered within a context. This cannot be extended/modified later.
2/ is the plan to support sparse ranges that can be registered, but mapped later? This does not seem handled correctly. And IMHO this is not needed.

io.c
```diff
@@ -2244,6 +2237,166 @@ static int io_sqe_files_register(struct io_ring_ctx *ctx, void __user *arg,
...
+
+	if (arg.max_len < arg.len) {
+		pr_err("%s: max len less than len", __func__);
+		goto err;
+	}
+
+	for (buf_index = 0; buf_index < PXD_IO_MAX_USER_BUFS; ++buf_index) {
+		imu = &ctx->user_bufs[buf_index];
+		if (imu->ubuf <= (uintptr_t) arg.base &&
+		    (uintptr_t) arg.base + arg.len <= imu->ubuf + imu->len) {
```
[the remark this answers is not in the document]
maxkozlovsky: added check

io.c
```diff
@@ -2244,6 +2237,166 @@ static int io_sqe_files_register(struct io_ring_ctx *ctx, void __user *arg,
...
+			offset = ((uintptr_t)arg.base - imu->ubuf) >> PAGE_SHIFT;
+			pr_info("found existing mapping %d, off %ld", buf_index, offset);
+			break;
+		}
+	}
+
+	if (buf_index == PXD_IO_MAX_USER_BUFS) {
```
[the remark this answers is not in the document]
maxkozlovsky: makes it harder to read this way for me

io.c
```diff
@@ -2244,6 +2237,166 @@ static int io_sqe_files_register(struct io_ring_ctx *ctx, void __user *arg,
...
+
+	imu = &ctx->user_bufs[buf_index];
+
+	/*
+         * Don't impose further limits on the size and buffer
+         * constraints here, we'll -EINVAL later when IO is
+         * submitted if they are wrong.
+         */
+	ret = -EFAULT;
+	if (!arg.base || !arg.len)
```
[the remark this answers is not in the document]
maxkozlovsky: does not matter much as the call is expected to succeed 100% of the time.

io.c
maxkozlovsky: added check

it allows for multiple set of buffers just the same, only there is a requirement to be done in one shot. This implementation relaxes that requirement.

io.c
```diff
@@ -2244,6 +2237,184 @@ static int io_sqe_files_register(struct io_ring_ctx *ctx, void __user *arg,
...
+	if (ret) {
+		/*
+                 * if we did partial map release any pages we did get
+                 */
+		if (pret > 0) {
+			for (j = 0; j < pret; j++)
+				put_page(pages[j]);
+		}
+		kvfree(imu->bvec);
+		imu->bvec = NULL;
```
sulakshm, line 2385: The above logic is incorrect. If there is any partial map failure, then in this attempt there was a part map failure. But the bvec may still hold pinned references to other successful mapping.

io.c
```diff
@@ -2244,6 +2237,184 @@ static int io_sqe_files_register(struct io_ring_ctx *ctx, void __user *arg,
...
+
+	ubuf = (unsigned long) arg.base;
+	end = (ubuf + arg.len) >> PAGE_SHIFT;
+	start = ubuf >> PAGE_SHIFT;
+	nr_pages = end - start;
+	max_nr_pages = arg.max_len >> PAGE_SHIFT;
+
+	if (existing) {
+		for (j = 0; j < nr_pages; ++j) {
+			if (imu->bvec[j + offset].bv_page) {
```
sulakshm, line 2333: imu->bvec assumed to be available if existing.
But if there is any map failure, the entire bvec is freed. So it opens up possibility for bvec to be NULL, but the mapping range to exist.

sulakshm (changes requested):

io.c
maxkozlovsky: added check for existing

io.c
```diff
@@ -2244,6 +2237,186 @@ static int io_sqe_files_register(struct io_ring_ctx *ctx, void __user *arg,
...
+		pr_err("%s: max len less than len", __func__);
+		goto err;
+	}
+
+	for (buf_index = 0; buf_index < PXD_IO_MAX_USER_BUFS; ++buf_index) {
+		imu = &ctx->user_bufs[buf_index];
+		if (imu->ubuf <= (uintptr_t) arg.base &&
+		    (uintptr_t) arg.base + arg.len <= imu->ubuf + imu->len) {
+			offset = ((uintptr_t)arg.base - imu->ubuf) >> PAGE_SHIFT;
+			pr_info("found existing mapping %d, off %ld", buf_index, offset);
```
sulakshm, line 2285: if one registers a range <buf,len,max_len> that falls within an existing map, max_len arg is not checked to be within the same map; only the passed active buf range<addr,len> is checked and passes silently, but max_len also should fall within the same map, otherwise should not be accepted as a pass. 

I fail to understand, why support sparse range at all? At any point in time, caller goes through the same ioctl, with the newer args. It would be just plain efficient and simpler, if a new range needs to be  registered to call the ioctl to register the range. In any case, code has been extended to allow multiple maps to be registered at any point in time. max_len is superfluous and should be removed.

Maybe ranges can be coalesced internally, but that does not buy any advantage at all.

maxkozlovsky: Could you please show me the code which demonstrates how the ranges will be easily coalesced? If this is too much trouble for you let's proceed with the current implementation. You only comment is that interface is inconsistent. This does not seem like a show stopper to me. You can always make changes that will improve the interface later.

sulakshm:
Based on how it's currently used from userspace where a larger 4G is
mapped, and smaller 32M is pinned in increasing order without unpin,
shouldn't the logic be similar to the loop that we have already.

wouldn't the below logic work. The passed address is definitely linear and
contiguous given its a single mmap request that is split.

for (buf_index = 0; buf_index < PXD_IO_MAX_USER_BUFS; ++buf_index) {
+		imu = &ctx->user_bufs[buf_index];
+		if (imu->ubuf + imu->len == arg.base) { // coalesced range here,
found = true

                    // if range is expanded more than limit (4G) then
fail the new mapping.

                          imu->len += arg.len;

                    }

 }

if none found, a new range gets created.


Regards

LN

maxkozlovsky:
You have omitted the coalescing part once again so it is still not clear what are you suggesting. I am guessing you want to reserve 4Gb regions all the time. This does not look like a good idea since this is going to waste memory if the caller has smaller mapping. Could we please move on? You can always fix it later.

sulakshm:
(imu->ubuf, imu->len) now maps to a larger range (imu->ubuf, imu->len)
to cover the new grow() method from userspace that registes additional
buffer.
The pinning remains the same in kernel.
And the index will still fall into the same range since the range got
coalesced as above.

I do not know how to make it more clear.

LN

maxkozlovsky:
Could you please post the code? This is not clear at all.

sulakshm:
```

struct io_mapped_ubuf {
   u64 ubuf;
   size_t len;
   struct bio_vec *bvec;
   unsigned int nr_bvecs;
};

u64 end_range(struct io_mapped_ubuf *ubuf)
{
    return ubuf->ubuf + ubuf->len;
}

struct pxd_ioc_register_buffers {
   void *base;
   size_t len;
};

// if below  needs to be a dynamic tunable, then have the max range part of the struct io_mapped_ubuf (it already is as nr_bvecs)
// allow a io ring context init to take a parameter to initialize this max range.
#define MAX_MAPPED_RANGE (1ULL << 32)


bool can_merge(struct io_mapped_ubuff *ubuff, struct pxd_ioc_register_buffers *arg)
{
    return (end_range(ubuff) == arg->base);
}

bool overlimit(struct io_mapped_ubuff *ubuff, struct pxd_ioc_register_buffers *arg)
{
   assert(can_merge(ubuff, arg));
   return (ubuff->len + arg->len > MAX_MAPPED_RANGE);
}

static int io_sqe_register_buffers(struct io_ring_ctx *ctx, void __user *uarg)
{
struct pxd_ioc_register_buffers arg;

copy_from_user(&arg, uarg, sizeof(arg))

// sanity check received arg.
if (!sanitize_args(arg))
   return -EINVAL;

// check for coalescing range.
for (buf_index=0; buf_index < PXD_IO_MAX_USER_BUFS; ++buf_index) {
  imu = &ctx->user_bufs[buf_index];
  if (can_merge(imu, &arg) && !overlimit(imu, &arg)) {
      imu->len += arg.len;
	  offset = ((uintptr_t)arg.base - imu->ubuf) >> PAGE_SHIFT;
	  pr_info("found existing mapping %d, off %ld", buf_index, offset);
	  break;
  }
  // sanitize for intersecting mapping.
  if (imu->ubuf < (uintptr_t) arg.base + arg.len &&
      (uintptr_t) arg.base < imu->ubuf + imu->len) {
			pr_info("%s: intersects with existing mapping", __func__);
			ret = -EEXIST;
			goto err;
  }
}

// did not find one range to coalesce current request, check for a new range.
if (buf_index == PXD_IO_MAX_USER_BUFS) {
		existing = false;
		for (buf_index = 0; buf_index < PXD_IO_MAX_USER_BUFS; ++buf_index) {
			pr_info("new mapping %d, ", buf_index);
			if (ctx->user_bufs[buf_index].ubuf == 0)
				break;
		}
		
}

if (buf_index == PXD_IO_MAX_USER_BUFS) {
		pr_err("%s: all busy", __func__);
		return -EBUSY;
	}

// this found 'buf_index' remains the same for new and merged reqs.
imu = &ctx->user_bufs[buf_index];


// passed new range
	ubuf = (unsigned long) arg.base;
	end = (ubuf + arg.len) >> PAGE_SHIFT;
	start = ubuf >> PAGE_SHIFT;
	nr_pages = end - start;
	max_nr_pages = MAX_MAPPED_RANGE >> PAGE_SHIFT;


// another sanity check that confirms no bvec pinned in the range.
if (existing) {
		for (j = 0; j < nr_pages; ++j) {
			if (imu->bvec[j + offset].bv_page) {
				pr_err("%s: trying to overwrite existing mapping", __func__);
				ret = -EEXIST;
				goto err;
			}
		}
	}
	
// pin range.
// assign bvec into the imu->bvec.

// this 'buf_index' is the expected result same for both merged and new range.
return buf_index;
}
```

Also, if unregister buffers are not supported, then remove the interface.
If it is supported, then please ensure it works.

sulakshm (approved):

PR #227, opened 2021-09-05
PWX-21175: handle kernel crash in fastpath + compilation fix

sulakshm:
**What this PR does / why we need it**:

In 2.10.0 branch

**Which issue(s) this PR fixes** (optional)
Closes #
https://portworx.atlassian.net/browse/PWX-21201 - fix compilation issue
https://portworx.atlassian.net/browse/PWX-21175 - fix kernel crash

**Special notes for your reviewer**:
This change passes the automation test consistently.
Fixes two issues.
1/ kernel crash
2/ compilation issue.

PASSED last 6 consecutive runs in the automation test.

PR #229, opened 2021-09-05
PWX-21175: handle kernel crash in fastpath

sulakshm:

Closes # https://portworx.atlassian.net/browse/PWX-21175

PASSED last 6 consecutive runs in the automation test.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1057,7 +1057,7 @@ static void pxd_rq_fn(struct request_queue *q)
 		fp_root_context_init(fproot);
 		if (pxd_dev->fp.fastpath) {
 			// route through fastpath
-			queue_work(pxd_dev->fp.wq, &fproot->work);
+			queue_work(fastpath_workqueue(), &fproot->work);
 			spin_lock_irq(&pxd_dev->qlock);
 			continue;
 		}
@@ -1133,7 +1133,7 @@ static blk_status_t pxd_queue_rq(struct blk_mq_hw_ctx *hctx,
 		// route through fastpath
 		// while in blkmq mode: cannot directly process IO from this thread... involves
 		// recursive BIO submission to the backing devices, causing deadlock.
-		queue_work(pxd_dev->fp.wq, &fproot->work);
+		queue_work(fastpath_workqueue(), &fproot->work);
 		return BLK_STS_OK;
 	}
 #endif
--- a/pxd.c
+++ b/pxd.c
@@ -1053,6 +1053,7 @@ static void pxd_rq_fn(struct request_queue *q)
 		fproot = &req->fproot;
 		fp_root_context_init(fproot);
 
+#ifdef __PX_FASTPATH__
 		if (pxd_dev->fp.fastpath) {
 			// route through fastpath
 			INIT_WORK(&fproot->work, fp_handle_io);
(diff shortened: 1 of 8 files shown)


pxd.c
```diff
@@ -1532,16 +1531,20 @@ ssize_t pxd_remove(struct fuse_conn *fc, struct pxd_remove_out *remove)
...
 	if (pxd_dev->disk && pxd_dev->disk->queue){
+#ifndef __PX_BLKMQ__
 		mutex_lock(&pxd_dev->disk->queue->sysfs_lock);
 
 		QUEUE_FLAG_SET(QUEUE_FLAG_DYING, pxd_dev->disk->queue);
 
-        mutex_unlock(&pxd_dev->disk->queue->sysfs_lock);
+		mutex_unlock(&pxd_dev->disk->queue->sysfs_lock);
+#else
+		blk_set_queue_dying(pxd_dev->disk->queue);
```
prabirpaul, line 1541: Why does QUEUE_FLAG_SET() not work for all cases?

pxd_bio_blkmq.c
```diff
@@ -580,19 +578,8 @@ static void fp_handle_specialops(struct work_struct *work) {
...
-                BUG_ON("unexpected condition");
-        }
-#else
-        if (REQ_OP(rq) & REQ_DISCARD) {
-                atomic_inc(&pxd_dev->fp.nio_discard);
-        } else {
-                BUG_ON("unexpected condition");
-        }
-#endif
+        BUG_ON(!rq_is_special(rq));
```
prabirpaul, line 581: rq_is_special() only checks for discard, Based on the below it should also check for WRITE_SAME and zeroout if there's an op for it.

prabirpaul (commented):
Could you please add a description for the issue being fixed.

sulakshm:
> Could you please add a description for the issue being fixed.

Please see detailed inline comments on the first PR https://github.com/portworx/px-fuse/pull/227

pxd.c
sulakshm: It may. There are additional actions done while marking a device queue as removing, please look inside blk_set_queue_dying. It would be safe to use a direct exported API instead.

pxd_bio_blkmq.c
sulakshm: No. px device only supports read/write and discard, write same and write zeros are not supported on px device.
The below logic converts the received discard to map to the backing device (btrfs file or dmthin volume).
It may convert discard internally to write same or write zero if discard is not available.

pxd_bio_blkmq.c
```diff
@@ -256,7 +257,9 @@ static int prep_root_bio(struct fp_root_context *fproot) {
                 return 0;
         }
 
-        rq_for_each_segment(bv, rq, rq_iter) nr_bvec++;
+        if (!specialops)
+                rq_for_each_segment(bv, rq, rq_iter) nr_bvec++;
+
```
prabirpaul, line 262: This is fine. But we need a take a closer look at it.
Since the compilation doesn't fail a kernel upgrade on an existing px node will result in a kernel panic and likely in a loop. We either not use apis like this or have a way of sanity checking the arguments before the call.

prabirpaul (approved):

pxd.c
sulakshm: PX_BLKMQ is defined from kernel 4.18+ or if el8 is defined. And in vanilla kernel is available from 3.19. We should be safe with this change. Will look out for any compilation issues after merge.

pxd_bio_blkmq.c
sulakshm: code cannot be future proof, since our driver is outside tree. I think the only focus we can make is issue these api calls only on a need basis and not assume they will behave the same otherwise. Like in this example, if this is a discard request, we do know there aren't any iovec carried, so skipping the call, but for read/writes this has to exist.
Yes exactly the latter from your argument, check if discard, then avoid making the call.

PR #251, opened 2022-05-02
PWX-23492 remove async list processing

maxkozlovsky:
Async list processing is not necessary since merging is done in the user
space, remove all the relevant code.

Closes #

### DIFF
--- a/io.c
+++ b/io.c
@@ -242,8 +242,6 @@ static size_t queue_size(struct io_ring_ctx *ctx)
 
 static int io_ring_ctx_init(struct io_ring_ctx *ctx, struct io_uring_params *params)
 {
-	int i;
-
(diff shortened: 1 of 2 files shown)


erichao1234 (approved):

erichao1234 (approved):
Just for my understanding - what do we use async lists for?

maxkozlovsky:
> Just for my understanding - what do we use async lists for?

That was some code which tried to merge the sequential requests in the kernel. The merge would only happen if the requests are truly sequential, that is there is a single stream, and the next request have not yet started processing by the workqueue. The success rate would be very small and can't be relied upon. Since the merging of the sequential requests is done in the user space, the kernel code duplicated the functionality (poorly) and pretty much wasted the cpu time and energy of the people who try to read the code.

PR #253, opened 2022-05-17
Unfreeze q after export

prabirpaul:
Pxd block device is exported out using add_disk(). It looks like in some cases while we are in the middle on this lsblk or systemd-udevd starts using the device. And the add_disk() gets stuck because device is in use.So to avoid that keep the device q frozen until after add_disk() returns so that add_disk always finishes. 

Closes #

Issue is not readily reproducible.
To test: 
1. Create a repl2 volumes on node1, node2.
2. Start 'watch -n 1 lsblk' on both nodes.
3. Start 'watch pxctl host attach' on both nodes.
Device attachment will flip-flop among the 2 nodes with lsblk continuously trying to read from device. No issue so far.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1255,6 +1255,10 @@ static int pxd_init_disk(struct pxd_device *pxd_dev, struct pxd_add_ext_out *add
 	q->queuedata = pxd_dev;
 	pxd_dev->disk = disk;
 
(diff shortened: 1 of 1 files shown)


sulakshm (approved):
ok.

PR #266, opened 2022-08-30
cleanup device remove/export path for race conditions

sulakshm:
1/ pxd_remove has a race where it leaves block device queue context hanging while context are deleted.
This allows new IOs to take a long on an invalid/freed up context and this locks up the system.

2/ pxd_add - revised to only update metadata in the driver. No kernel side exposure gets made.

3/ pxd_export - revised to only export to kernel with the corresponding pxd device info.

Closes #

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -65,9 +65,6 @@ static DEFINE_IDA(pxd_minor_ida);
 
 static DEFINE_MUTEX(pxd_ctl_mutex);
 
-#define MINOR_GET()  ida_simple_get(&pxd_minor_ida, 1, 1 << MINORBITS, GFP_KERNEL)
-#define MINOR_PUT(m) ida_simple_remove(&pxd_minor_ida, m)
-
 struct pxd_context *pxd_contexts;
 uint32_t pxd_num_contexts = PXD_NUM_CONTEXTS;
 uint32_t pxd_num_contexts_exported = PXD_NUM_CONTEXT_EXPORTED;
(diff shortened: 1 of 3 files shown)


pxd.c
```diff
@@ -63,6 +63,11 @@ extern const char *gitversion;
 static dev_t pxd_major;
 static DEFINE_IDA(pxd_minor_ida);
 
+static DEFINE_MUTEX(pxd_ctl_mutex);
+
+#define MINOR_GET()  ida_simple_get(&pxd_minor_ida, 1, 1 << MINORBITS, GFP_KERNEL)
+#define MINOR_PUT(m) ida_simple_remove(&pxd_minor_ida, m)
```
prabirpaul, line 69: These 2 aren't used anywhere?

pxd.c
```diff
@@ -1520,24 +1592,38 @@ ssize_t pxd_remove(struct fuse_conn *fc, struct pxd_remove_out *remove)
 	pr_info("removing device %llu", pxd_dev->dev_id);
 
 	/* Make sure the req_fn isn't called anymore even if the device hangs around */
-	if (pxd_dev->disk && pxd_dev->disk->queue){
+	if (pxd_dev->exported && pxd_dev->disk && pxd_dev->disk->queue){
```
prabirpaul, line 1595: If the queue exists, it should be marked dying. Doesn't matter exported is set or not.

pxd.c
```diff
@@ -1520,24 +1592,38 @@ ssize_t pxd_remove(struct fuse_conn *fc, struct pxd_remove_out *remove)
...
 #endif
 	}
 
+	cleanup = pxd_dev->exported;
+	pxd_dev->exported = false;
 	spin_unlock(&pxd_dev->lock);
 
-	pxd_fastpath_reset_device(pxd_dev);
-	device_unregister(&pxd_dev->dev);
+	if (cleanup) {
```
prabirpaul, line 1616: Why is the exported check needed? It's not going to protect against concurrent pxd_remove() on the same device.

pxd.c
```diff
@@ -63,6 +63,11 @@ extern const char *gitversion;
 static dev_t pxd_major;
 static DEFINE_IDA(pxd_minor_ida);
 
+static DEFINE_MUTEX(pxd_ctl_mutex);
```
prabirpaul, line 66: Don't we need to destroy this at module exit?

sulakshm: nope.. thats a module local... nothing to destroy

pxd.c
sulakshm: removed

pxd.c
sulakshm: exported is set to true only if disk and disk queue both are setup properly.
If during setup, either of them fail, then it is all cleaned up there, so one cannot have disk without queue.
The added check should not hurt.

pxd.c
sulakshm: If pxd_remove is run unconditionally without looking at exported flag, there is an assumption that from userspace pxd_remove gets called only if device add succeeds and is exported to the kernel.

If for some reason, only pxd_add got called and no followup pxd_export was called, but pxd_remove now follows from userspace, then it will likely crash the kernel. This though currently does not happen.
So having the check doesnt hurt.

prabirpaul (approved):

PR #268, opened 2022-09-13
Set pxd minors to 1, pxd disk can't be partitioned

prabirpaul:
It's incorrect to set minors to 256 since alloc_disk() was called with minors==1.

Closes #

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1261,7 +1261,7 @@ static int pxd_init_disk(struct pxd_device *pxd_dev, struct pxd_add_ext_out *add
 	snprintf(disk->disk_name, sizeof(disk->disk_name),
 		 PXD_DEV"%llu", pxd_dev->dev_id);
 	disk->major = pxd_dev->major;
-	disk->minors = 256;
+	disk->minors = 1;
 	disk->first_minor = pxd_dev->minor;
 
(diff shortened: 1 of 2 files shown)


pxd.c
```diff
@@ -1261,7 +1261,7 @@ static int pxd_init_disk(struct pxd_device *pxd_dev, struct pxd_add_ext_out *add
 	snprintf(disk->disk_name, sizeof(disk->disk_name),
 		 PXD_DEV"%llu", pxd_dev->dev_id);
 	disk->major = pxd_dev->major;
-	disk->minors = 256;
+	disk->minors = 1;
```
sulakshm, line 1264: No, this should not be set by us. It is initialized part of alloc_disk argument.

sulakshm:
Would it be possible to put pxd_suspend_io() also same as in 2.10.0? This is in pxd_bio_blkmq.c

pxd.c
prabirpaul: Kernel 5.15 onwards this is required since alloc_disk doesn't do it. Will move it up.

prabirpaul:
cherry-picked origin/ln/nominors

sulakshm (approved):

PR #296, opened 2023-10-27
PWX-31637: set iterator direction based on kernel version. Prevent older kernels builds from breaking.

jrivera-px:

Closes #
PWX-31637

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1260,6 +1260,7 @@ static int pxd_init_disk(struct pxd_device *pxd_dev, struct pxd_add_ext_out *add
 		 PXD_DEV"%llu", pxd_dev->dev_id);
 	disk->major = pxd_dev->major;
 	disk->first_minor = pxd_dev->minor;
+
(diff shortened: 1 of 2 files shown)


pxd.c
```diff
@@ -181,8 +181,13 @@ static long pxd_ioctl_init(struct file *file, void __user *argp)
...
 	struct iov_iter iter;
 	struct iovec iov = {argp, sizeof(struct pxd_ioctl_init_args)};
+	int direction = WRITE;
 
-	iov_iter_init(&iter, READ, &iov, 1, sizeof(struct pxd_ioctl_init_args));
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6,2,0)
+	direction = READ;
+#endif
+	  
+	iov_iter_init(&iter, direction, &iov, 1, sizeof(struct pxd_ioctl_init_args));
```
sulakshm, line 190: direction is either ITER_SOURCE (for write) or ITER_DEST (for read).
Using READ or WRITE is incorrect here.

PR #298, opened 2023-11-30
PWX-35199: Null pointer crash fix

sulakshm:
As part of vsphere dmthin testing observed kernel crash.
With changes from this branch and PR the crash is not observed anymore.

Closes # PWX-35199


prabirpaul (approved):
Should be required in master as well.

PR #299, opened 2023-12-14
harden device remove

sulakshm:
1/ harden device removal path
--- freeze queue and mark device dead to ensure newer IOs do not continue to flow while device is being detached.

2/ fix the version report that does not account for the null char.

Closes # https://portworx.atlassian.net/browse/PWX-35419

1/ harden volume detach path

QA note: 
posted changes are running for 2 days with the issue not seen.

2/ also fix version report include null
```
// exact match
#cat px_version.c
const char *gitversion = "ln/minordevs:34234325345435345464565453524234242432423423423477";
# ./pxd -v
Version: ln/minordevs:34234325345435345464565453524234242432423423423477
#

// more than 64
#cat px_version.c
const char *gitversion = "ln/LARGER:34234325345435345464565453524234242432423423423477888888888888";
# ./pxd -v
Version: ln/LARGER:34234325345435345464565453524234242432423423423477888
#


// less than 64
# cat px_version.c
const char *gitversion = "ln/SMALL:2345454";
# ./pxd -v
Version: ln/SMALL:2345454
#
```

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1615,8 +1615,8 @@ ssize_t pxd_remove(struct fuse_conn *fc, struct pxd_remove_out *remove)
 		mutex_unlock(&pxd_dev->disk->queue->sysfs_lock);
 #else
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(5,13,0)
-		// Do not mark queue dead, del_gendisk will try to
-		// submit all outstanding IOs on this device
+		blk_freeze_queue_start(pxd_dev->disk->queue);
+		blk_mark_disk_dead(pxd_dev->disk);
 #else
 		blk_set_queue_dying(pxd_dev->disk->queue);
 #endif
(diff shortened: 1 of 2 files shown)


pxd.c
```diff
@@ -1615,8 +1619,8 @@ ssize_t pxd_remove(struct fuse_conn *fc, struct pxd_remove_out *remove)
 		mutex_unlock(&pxd_dev->disk->queue->sysfs_lock);
 #else
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(5,13,0)
-		// Do not mark queue dead, del_gendisk will try to
-		// submit all outstanding IOs on this device
+		blk_freeze_queue_start(pxd_dev->disk->queue);
+		blk_mark_disk_dead(pxd_dev->disk);
```
prabirpaul, line 1623: del_gendisk is still getting called from this path
this can be added only after we have removed the pxd_free_disk() below

sulakshm: the above are calls needed for handling surprise device removal case.
i haven't merged the async detach changes which we discussed offline.

prabirpaul (approved):

PR #303, opened 2024-01-11
PWX-35356: Fix compilation issues for 6.5.12-100.fc37.x86_64

jrivera-px:
**What this PR does / why we need it**

Support for 6.5 kernels. 

Closes #
PWX-35356

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -90,12 +90,19 @@ struct pxd_context* find_context(unsigned ctx)
 	return &pxd_contexts[ctx];
 }
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6,5,0)
+static int pxd_open(struct gendisk *bdev, blk_mode_t mode)
+#else
 static int pxd_open(struct block_device *bdev, fmode_t mode)
-{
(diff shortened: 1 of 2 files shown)


pxd.c
```diff
@@ -132,8 +143,11 @@ static void pxd_release(struct gendisk *disk, fmode_t mode)
 	BUG_ON(pxd_dev->magic != PXD_DEV_MAGIC);
 	pxd_dev->open_count--;
 	spin_unlock(&pxd_dev->lock);
-
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6,5,0)
+	trace_pxd_release(pxd_dev->dev_id, pxd_dev->major, pxd_dev->minor);
```
prabirpaul, line 147: Or just get rid of it. I doubt this is ever going to be used.

prabirpaul (approved):

pxd.c
jrivera-px: True,  not clear when tracing is enabled or how we enable it so worked around it...  thx

PR #306, opened 2024-02-19
PWX-35753: handle device detach timeouts 

sulakshm:
handle device detach timeouts and subsequent attach incorrectly passing because of stale block device. 

Closes # PWX-35753

https://jenkins.pwx.dev.purestorage.com/job/Dev/job/Porx-08/116
```
block device detached
block device detachment scheduled and wait upon with a timeout
detachment action timeout
px-storage and px continues further with block device marked detached, while block device detachment still continues in the kernel block layer
new request to attach same block device
block device gets attached with a nop from the kernel fuse driver because of stale device that is being detached.
block attachment fails with no block device,  because previous stale device got removed.
```

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -215,6 +215,7 @@ static long pxd_ioctl_fp_cleanup(struct file *file, void __user *argp)
 	struct pxd_fastpath_out cleanup_args;
 	long ret = 0;
 	struct pxd_device *pxd_dev;
+	bool stale;
 
 	if (copy_from_user(&cleanup_args, argp, sizeof(cleanup_args))) {
 		return -EFAULT;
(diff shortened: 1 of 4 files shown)


pxd.c
```diff
@@ -1323,15 +1329,17 @@ static void pxd_free_disk(struct pxd_device *pxd_dev)
...
 {
 	struct pxd_device *pxd_dev_itr, *pxd_dev;
 
 	pxd_dev = NULL;
+	*stale = false;
 	spin_lock(&ctx->lock);
 	list_for_each_entry(pxd_dev_itr, &ctx->list, node) {
 		if (pxd_dev_itr->dev_id == dev_id) {
 			pxd_dev = pxd_dev_itr;
+			*stale = pxd_dev->removing;
```
adikulkarni0, line 1342: The ctx->lock protects the list, do we need to the hold the lock when removing is set or checked? 
It looks like we grab the pxd_dev->lock while setting this instead. Although its grabbed with the ctx->lock held too.
I dont think the code change is wrong but is confusing as there seems to be implicit assumption that when ctx->lock is help the pxd_dev->lock is also held when updating removing. 
If possible can you add some comments to the code where these both locks are defined so that we know what these locks are supposed to protect?

adikulkarni0: can we fix this implicit assumption of the lock and grab the pxd_dev->lock when we look at removing field?

adikulkarni0 (approved):
Other than the 1 open comment nothing major. 
For the new comments added can you check if they are going beyond the 80 char limit per line as per the coding style guide. (I am not sure if we follow the linux kernel coding style guide)

pxd.c
vivekpx: added lock for checking `removing`

PR #308, opened 2024-03-09
PWX-36427: do not call add_disk with spinlock held

sulakshm:
BUG: scheduling while atomic: px-storage/15156/0x00000002 seen during pool expand and node restart
avoid this issue.

Closes # https://portworx.atlassian.net/browse/PWX-36427

This is already addressed in master branch.
Please ensure a new base branch gets created for release before merging.
This PR is on top of v3.1.0 px-fuse driver.

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1580,6 +1580,7 @@ ssize_t pxd_export(struct fuse_conn *fc, uint64_t dev_id)
 		goto cleanup;
 	}
 
+	spin_unlock(&pxd_dev->lock);
 	err = pxd_bus_add_dev(pxd_dev);
 	if (err) {
 		pxd_free_disk(pxd_dev);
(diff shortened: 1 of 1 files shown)


pxd.c
```diff
@@ -1580,29 +1580,32 @@ ssize_t pxd_export(struct fuse_conn *fc, uint64_t dev_id)
 		goto cleanup;
 	}
 
+	spin_unlock(&pxd_dev->lock);
```
adikulkarni0, line 1583: Couple of issues here. 
1. We set the pxd->exported with the spinlock held, but is checked without grabbing the lock. This is existing code, not your changes.
2. With this change the calls like  pxd_bus_add_dev, device_add_disk can happen in parallel for the same device, which previously didnt happen.

I recommend below approaches to avoid the sleep with the spinlock held:
1. Consider a state in the pxd_dev which indicates that the device is undergoing the process of exporting. Iff the current exported bool is not enough.
The code will now look as below:
```
pxd_export() {
     spin_lock(&pxd_dev->lock);
     if (pxd_dev->exporting) {
           return -EBUSY;
     }
     pxd_dev->exporting = true;
     ...
     spin_unlock(&pxd_dev->lock);
     err = pxd_bus_add_dev(pxd_dev);
     err = device_add_disk(&pxd_dev->dev, pxd_dev->disk, NULL);
     spin_lock(&pxd_dev->lock);
     pxd_dev->exported = true;
     spin_unlock(&pxd_dev->lock);
     ...
}
```
Use atomic if it makes more sense here.

sulakshm:
closed as not reviewed on time for this branch

PR #314, opened 2024-05-22
PWX-37390: revise discard granularity for pxd devices

sulakshm:
pxd devices currently hard code discard granularity which gets assigned to pxd device.
if those pxd devices receives discard which are not aligned to advertised real target devices, then userspace process can intelligently split and act correctly.
But while in fastpath, if the requests are not aligned to the actual device attribute then (some or all) discards become nop.
There currently exist no interface from userspace to configure this value dynamically.
For now, revising the discard param to the max supported value at the raw device end (which would be 2M),

Test logs are attached part of https://github.com/pure-px/porx/pull/13439/
Closes # PWX-37390

### DIFF
--- a/pxd.h
+++ b/pxd.h
@@ -76,7 +76,7 @@ struct pxd_ioc_free_buffers {
 #define PXD_MAX_IO		(1024*1024)	/**< maximum io size in bytes */
 #define PXD_MAX_QDEPTH  256			/**< maximum device queue depth */
 #define PXD_MIN_DISCARD_GRANULARITY		PXD_LBS
-#define PXD_MAX_DISCARD_GRANULARITY		(64 * 1024)
+#define PXD_MAX_DISCARD_GRANULARITY		(2 << 20)
 
 // NOTE: nvme devices can go upto 1023 queue depth
 #define MAX_CONGESTION_THRESHOLD (1024)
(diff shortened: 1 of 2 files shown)


pxd.h
```diff
@@ -76,7 +76,7 @@ struct pxd_ioc_free_buffers {
 #define PXD_MAX_IO		(1024*1024)	/**< maximum io size in bytes */
 #define PXD_MAX_QDEPTH  256			/**< maximum device queue depth */
 #define PXD_MIN_DISCARD_GRANULARITY		PXD_LBS
-#define PXD_MAX_DISCARD_GRANULARITY		(64 * 1024)
+#define PXD_MAX_DISCARD_GRANULARITY		(2 << 20)
```
sangleganesh, line 79: will this affect the Portworx processes which don't handle large discards ?

sulakshm:
Ganesh,
This change implies block device will only support discard granularity at
2M which initially was at 64K.
Filesystems on top honour this granularity correctly, but  the impact of
any performance op needs to be understood.
I can't definitely tell whether fs will be severely affected or wont see
any difference at all, as now.
Theoretically, this change, should fix fs from incorrectly submitting
discards (on pxd devices) which becomes nop on the block devices (replica
targets) if the  alignments do not match up.

Wayne has been requested to have a run to understand this behavior.
Hopefully it brings some clarity to this further.
LN

sangleganesh (approved):

PR #334, opened 2025-03-19
Pwx 42778 fix

sreymondjohn:
Revert changes for returning -EFAULT on user iovec count reaching 0 but there is data to be copied from bvecs since it is a valid scenario for aggr volumes.

Closes #


sangleganesh (approved):
let us add a comment in the commit that the user space may split the request and make mulitple read calls, so this check needs to be reverted.

PR #345, opened 2025-04-07
PWX-42981

sreymondjohn:
Adds the following traces

```
fastpath request/reply
nativepath request/reply, along with copying data from px-fuse to px-storage
fallback start and end
failover start and end
reroute to native path (there are multiple places where this can happen)
```
Closes # PWX-42981

### DIFF
--- a/dev.c
+++ b/dev.c
@@ -578,6 +578,7 @@ static int fuse_notify_add_ext(struct fuse_conn *conn, unsigned int size,
 		printk(KERN_ERR "%s: can't copy arg\n", __func__);
 		return -EFAULT;
 	}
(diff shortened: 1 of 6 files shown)


sulakshm:
can you exercise each of those trace paths and add the logs here? 
will this get integrated with lttng-cmd cli control to enable tracing at driver level too?

sreymondjohn:
> can you exercise each of those trace paths and add the logs here? 

Sure, @sulakshm , I can add that, working on other tickets, will do this by end of week.

> will this get integrated with lttng-cmd cli control to enable tracing at driver level too?
that would be a good idea, once this goes in, I can create an improvement ticket in PWX.

sreymondjohn:
Closing this PR and will open a new PR against `main` with the debug logs

PR #349, opened 2025-04-08
PWX-43085: allow disabling discards

sulakshm:
allow disabling discards on pxd device.

Linked porx PR: https://github.com/pure-px/porx/pull/14751

Closes # PWX-43085

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1370,10 +1370,7 @@ static int pxd_init_disk(struct pxd_device *pxd_dev)
 
     q->limits.discard_granularity = PXD_MAX_DISCARD_GRANULARITY;
     q->limits.discard_alignment = PXD_MAX_DISCARD_GRANULARITY;
-	if (pxd_dev->discard_size < SECTOR_SIZE)
-		q->limits.max_discard_sectors = PXD_MAX_IO / SECTOR_SIZE;
-	else
-		q->limits.max_discard_sectors = pxd_dev->discard_size / SECTOR_SIZE;
+    q->limits.max_discard_sectors = pxd_dev->discard_size / SECTOR_SIZE;
 
 #if LINUX_VERSION_CODE < KERNEL_VERSION(4,12,0)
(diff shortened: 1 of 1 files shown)


pxd.c
```diff
@@ -1521,13 +1518,15 @@ ssize_t pxd_add(struct fuse_conn *fc, struct pxd_add_ext_out *add)
...
 
-	if (add->discard_size < SECTOR_SIZE)
-		pxd_dev->discard_size = PXD_MAX_IO;
+	if (add->discard_size == 0) {
+		pxd_dev->discard_size = 0; // disabled
+	} else if (add->discard_size <= PXD_MAX_DISCARD_GRANULARITY)
+		pxd_dev->discard_size = PXD_MAX_DISCARD_GRANULARITY; // unaligned make it minimum aligned
 	else
-		pxd_dev->discard_size = add->discard_size;
+		pxd_dev->discard_size = ALIGN(add->discard_size, PXD_MAX_DISCARD_GRANULARITY); // aligned
```
dgoel-px, line 1526: Iiuc this will change the default discard size from 1MB to 64k. Can you help understand why we need this change and what is the implication?

dgoel-px: On further reading, given that we only specify two discard sizes in PX (4k and 10*4k) the condition (add->discard_size <= PXD_MAX_DISCARD_GRANULARITY) will never be hit?

dgoel-px: On further reading, given that we only specify two discard sizes in PX (4k and 10*4k) the condition (add->discard_size <= PXD_MAX_DISCARD_GRANULARITY) will never be hit?

sreymondjohn (approved):

pxd.c
sulakshm: it shall be configurable and would want that to be atleast 2M. That would be in porx side. This piece only ensure settings on the block device are aligned and allows disabling it.

dgoel-px (approved):

PR #353, opened 2025-04-11
PWX-43070

sreymondjohn:
For reference, use https://docs.kernel.org/RCU/Design/Requirements/Requirements.html, and it has a specific example which is similar to our use-case

```
1 #define STATE_NORMAL        0
 2 #define STATE_WANT_RECOVERY 1
 3 #define STATE_RECOVERING    2
 4 #define STATE_WANT_NORMAL   3
 5
 6 int state = STATE_NORMAL;
 7
 8 void do_something_dlm(void)
 9 {
10   int state_snap;
11
12   rcu_read_lock();
13   state_snap = READ_ONCE(state);
14   if (state_snap == STATE_NORMAL)
15     do_something();
16   else
17     do_something_carefully();
18   rcu_read_unlock();
19 }
20
21 void start_recovery(void)
22 {
23   WRITE_ONCE(state, STATE_WANT_RECOVERY);
24   synchronize_rcu();
25   WRITE_ONCE(state, STATE_RECOVERING);
26   recovery();
27   WRITE_ONCE(state, STATE_WANT_NORMAL);
28   synchronize_rcu();
29   WRITE_ONCE(state, STATE_NORMAL);
30 }
```

```
The RCU read-side critical section in do_something_dlm() works with the synchronize_rcu() in start_recovery() to guarantee that do_something() never runs concurrently with recovery(), but with little or no synchronization overhead in do_something_dlm().
```

What this PR does / why we need it:
the race is describe in detail here : https://purestorage.atlassian.net/browse/PWX-43070

Which issue(s) this PR fixes (optional)
Closes # PWX-43070

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1185,6 +1185,11 @@ static blk_status_t pxd_queue_rq(struct blk_mq_hw_ctx *hctx,
 {
 	struct fp_root_context *fproot = &req->fproot;
 	fp_root_context_init(fproot);
+
+	// rcu read side critical section can't sleep
+	// fastpath_queue_work -> kthread_queue_work
+	// and kthread_queue_work won't sleep since it
+	// acquires a spinlock (see : https://elixir.bootlin.com/linux/v6.13.7/source/kernel/kthread.c#L1030)
 	rcu_read_lock();
 	if (READ_ONCE(pxd_dev->fp.fastpath)) {
 		// route through fastpath
--- a/pxd.c
+++ b/pxd.c
@@ -1185,13 +1185,16 @@ static blk_status_t pxd_queue_rq(struct blk_mq_hw_ctx *hctx,
 {
 	struct fp_root_context *fproot = &req->fproot;
 	fp_root_context_init(fproot);
-	if (pxd_dev->fp.fastpath) {
+	rcu_read_lock();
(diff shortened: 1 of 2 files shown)


sulakshm (approved):

pxd.c
```diff
@@ -1185,13 +1185,21 @@ static blk_status_t pxd_queue_rq(struct blk_mq_hw_ctx *hctx,
...
+	// fastpath_queue_work -> kthread_queue_work
+	// and kthread_queue_work won't sleep since it
+	// acquires a spinlock (see : https://elixir.bootlin.com/linux/v6.13.7/source/kernel/kthread.c#L1030)
+	rcu_read_lock();
+	if (READ_ONCE(pxd_dev->fp.fastpath)) {
 		// route through fastpath
 		// while in blkmq mode: cannot directly process IO from this thread... involves
 		// recursive BIO submission to the backing devices, causing deadlock.
 		fastpath_queue_work(&fproot->work, false);
+		rcu_read_unlock();
```
sulakshm, line 1199: The original race is condition and action is not atomic. Adding a lock around it made it atomic, so the race is gone.
The cost of acquiring read locks during an infrequent update should be understood so as not to affect performance.
BUT rcu is a perfect fit for infrequent updates as in this case, so this should be okay

PR #366, opened 2025-06-24
PWX-XXXX support discard on aligned writezero

hsahoo-px:

Closes #

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -791,6 +791,27 @@ static int pxd_discard_request(struct fuse_req *req, uint32_t size, uint64_t off
 	return 0;
 }
 
+static int pxd_write_zeroes_request(struct fuse_req *req, uint32_t size, uint64_t off,
+                                   uint32_t minor, uint32_t flags)
+{
+    int rc;
+	
+    // Handle device limits check
+    rc = pxd_handle_device_limits(req, &size, &off, REQ_OP_WRITE_ZEROES);
+    if (rc) {
+        return rc;
+    }
+
+    req->in.h.opcode = PXD_WRITE_ZEROES;  // New FUSE opcode
+#ifdef __PXD_BIO_MAKEREQ__
+    req->end = pxd_process_write_reply;
+#else
+    req->end = pxd_process_write_reply_q;
+#endif
+    pxd_req_misc(req, size, off, minor, flags);
+    return 0;
+}
+
 static int pxd_write_same_request(struct fuse_req *req, uint32_t size, uint64_t off,
 			uint32_t minor, uint32_t flags)
 {
@@ -824,15 +845,16 @@ static int pxd_request(struct fuse_req *req, uint32_t size, uint64_t off,
 {
 	int rc;
 	trace_pxd_request(req->in.h.unique, size, off, minor, flags);
-
 	switch (op) {
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,18,0) || (LINUX_VERSION_CODE >= KERNEL_VERSION(5,14,0) && (defined(__EL8__) || defined(__SUSE_EQ_SP5__)))
+// #if LINUX_VERSION_CODE >= KERNEL_VERSION(5,18,0) || (LINUX_VERSION_CODE >= KERNEL_VERSION(5,14,0) && (defined(__EL8__) || defined(__SUSE_EQ_SP5__)))
 	case REQ_OP_WRITE_ZEROES:
-#else
+		rc = pxd_write_zeroes_request(req, size, off, minor, flags);
+		break;
+// #else
 	case REQ_OP_WRITE_SAME:
-#endif
 		rc = pxd_write_same_request(req, size, off, minor, flags);
 		break;
+// #endif
 	case REQ_OP_WRITE:
 		rc = pxd_write_request(req, size, off, minor, flags);
 		break;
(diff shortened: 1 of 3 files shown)


pxd.h
```diff
@@ -86,6 +86,7 @@ enum pxd_opcode {
 	PXD_FALLBACK_TO_KERNEL,   /**< Fallback requests suspend IO and send in a marker req
 						  from kernel on a suspended device */
 	PXD_EXPORT_DEV,     /**< export the attached device to the kernel */
+	PXD_WRITE_ZEROES,   /**< write_zeroes operation */
```
sulakshm, line 89: need to add features export for new op code.
So px-storage knows that this driver supports write zeroes control.

pxd.h
```diff
@@ -97,6 +98,7 @@ enum pxd_opcode {
 
 #define PXD_LBS (4 * 1024) 	/**< logical block size */
 #define PXD_LBS_MASK (PXD_LBS - 1)
+#define PXD_MAX_WRITE_ZEROES_SIZE (2 * 1024 * 1024) /**< max write zeroes size in bytes */
```
sulakshm, line 101: since its a discard overloading internally, use same max PXD_MAX_DISCARD_GRANULARITY

dev.c
```diff
@@ -554,8 +558,41 @@ static int fuse_notify_add_ext(struct fuse_conn *conn, unsigned int size,
...
+	size_t len;
+
+	if (size < expected_size) {
+		printk(KERN_ERR "%s: invalid size %u (expected %zu)\n", __func__, size, expected_size);
+		return -EINVAL;
+	}
+
+	len = expected_size;
+	memset(&add, 0, sizeof(add));  // Zero-initialize to handle any padding
+
```
sulakshm: the logic above seems incorrect or not needed.
px-storage will use the full structure to pass this ioctl and it has to match its size below; i.e. len = sizeof(pxd_add_ext_out_v2)

also the structure can be shorter, pxd_add_v2_out instead?

dev.c
```diff
@@ -554,8 +558,41 @@ static int fuse_notify_add_ext(struct fuse_conn *conn, unsigned int size,
...
+	// Validate reserved field is zero (for future compatibility)
+	if (add.reserved != 0) {
+		printk(KERN_WARNING "%s: reserved field is non-zero (%llu), ignoring\n",
+		       __func__, add.reserved);
+	}
+
+	// Cast to pxd_add_ext_out for pxd_add() - the extra fields are at the end
 	trace_fuse_notify_add_ext(add.dev_id, add.size, add.queue_depth, add.discard_size, add.open_mode, add.enable_fp, add.paths.count);
-	return pxd_add(conn, &add);
+	return pxd_add(conn, (struct pxd_add_ext_out *)&add, true);  // v2 structure
```
sulakshm: This is wrong... pxd_add() should use pxd_add_v2_out structure as input and all existing interface should convert to this.

pxd_fastpath.c
```diff
@@ -559,6 +542,18 @@ void enableFastPath(struct pxd_device *pxd_dev, bool force)
...
 	pxd_dev->fp.fastpath = true;
+
+	// Disable WriteZero→Discard optimization for fastpath
+	// Fastpath uses LVM which doesn't support discard, so WriteZero must be disabled
+	// even if the device was previously in native path with WriteZero enabled
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4,10,0)
+	if (pxd_dev->disk && pxd_dev->disk->queue) {
+		blk_queue_max_write_zeroes_sectors(pxd_dev->disk->queue, 0);
+		printk(KERN_INFO"device %llu: disabled WriteZero for fastpath (LVM doesn't support discard)\n",
+		       pxd_dev->dev_id);
```
sulakshm: have you validated this path? does it work as you have coded?
Attach a device in native path with WZ enabled, then fallback to fastpath, check state, then failover and check WZ state on the same device.

pxd.c
```diff
@@ -830,6 +855,15 @@ int pxd_initiate_failover(struct pxd_device *pxd_dev)
 		return -ENODEV;
 	}
 
+	// Block fastpath failover if WriteZero→Discard optimization is enabled
+	// Fastpath (LVM) doesn't support discard, so WriteZero must remain disabled
+	// Allowing failover would cause data corruption (reads after discard return garbage)
+	if (pxd_dev->enable_write_zeroes_support) {
```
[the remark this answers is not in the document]
hsahoo-px: Added capabilities as suggested.

pxd_fastpath.c
hsahoo-px: Currently the behavior is kept as follow.
WZ will be enabled only for native path and fastPath it will be disabled..
If anyone moving from native to fastPath, as part of detach and reattach wz will be disabled again.
If anyone moving from fastpath to native path then the wz will be in disable state only. If the device will be reattached, then the wz will be enabled.

Currently the wz will be set only device creation and attach code flow.

pxd.h
```diff
@@ -159,6 +161,25 @@ struct pxd_add_ext_out {
 	struct pxd_update_path_out paths; /**< backing device paths */
 };
 
+/* Device capability flags */
+#define PXD_CAP_WRITE_ZEROES    (1 << 0)  /* WriteZero→Discard optimization */
```
sulakshm: nit.
this is a per device capability for write zeroes. Can you name it accordingly.
PXD_DEV_CAP_WRITE_ZEROES.

pxd_bio_blkmq.c
```diff
@@ -657,7 +657,7 @@ static void fp_handle_specialops(struct kthread_work *work) {
 	  r = blkdev_issue_zeroout(bdev, blk_rq_pos(rq),
 				   blk_rq_sectors(rq), GFP_NOIO, 0);
 	}
-#elif LINUX_VERSION_CODE >= KERNEL_VERSION(5,18,0) 
+#elif LINUX_VERSION_CODE >= KERNEL_VERSION(5,18,0)
```
sulakshm: no change to this file - remove this change out of PR

PR #379, opened 2025-07-29
PWX-45104 Kernel driver changes for dynamic threads

nikhil-bhupale:
Added sysfs parameter to updaet pxd_num_fpthreads. Adjust number of worker threads as per pxd_num_fpthreads


nikhil-bhupale:
Performance results are added to following ticket
https://purestorage.atlassian.net/browse/PWX-45395

pxd.c
```diff
@@ -1915,16 +1921,54 @@ static void pxd_root_dev_release(struct device *dev)
...
+static struct device_type pxd_root_device_type = {
+	.name		= "pxd",
+	.groups		= pxd_root_attr_groups,
+	.release	= pxd_root_dev_release,
+};
+
 static struct device pxd_root_dev = {
 	.init_name =    "pxd",
 	.release =      pxd_root_dev_release,
+	.type = &pxd_root_device_type,
```
sulakshm, line 1969: since you introduced at the top level to configure pxd devices - which is absolutely the correct thing to do...
unrelated but relevant:
can you bring pxd_timeout_show/store also here and control the pxd_timeout_msecs.
Currently this configuration is possible only if a pxd device exists through its minor dev path. - very ugly.

pxd_fastpath.c
```diff
@@ -70,6 +69,94 @@ static void fastpath_flush_work(void) {
...
+
+	printk(KERN_NOTICE"Updated fpthreads from %d to %d", pxd_num_fpthreads, new_pxd_num_fpthreads);
+	pxd_num_fpthreads = new_pxd_num_fpthreads;
+	return 0;
+
+out:
+	printk(KERN_ERR"fastpath fpthread init failure %d\n", rc);
+	for (node=0; node < MAX_NUMNODES; node++) {
+		struct pxfpcontext_per_node *c = &pxfpctxt[node];
+		if (c->valid) {
```
sulakshm, line 148: how is now c->valid used? if the number of fp threads change. i do not see it modified anywhere.

pxd_fastpath.c
```diff
@@ -70,6 +69,94 @@ static void fastpath_flush_work(void) {
...
+			// While decreasing active will reach new_pxd_num_fpthreads
+			// without creating new threads
+			if (c->fpworker[active] != NULL) {
+				active++;
+				continue;
+			}
+
+			// We need increase num threads, so create more threads
+			printk(KERN_NOTICE"Create thread on numa node %d cpu %d\n", node, cpu);
+			worker = kthread_create_worker_on_cpu(cpu, 0, "pxfpn%dc%d", node, cpu);
```
sulakshm, line 119: thread resources are never destroyed if decremented?

pxd_fastpath.c
```diff
@@ -26,11 +26,10 @@ static struct workqueue_struct *gwq;
 extern uint32_t pxd_num_fpthreads;
 
 #define MAX_PXFP_WORKERS_PER_NODE (pxd_num_fpthreads) /// keep it power of 2.
```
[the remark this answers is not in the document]
nikhil-bhupale: yes, I have removed check to block non power of 2, to allow more granularity

pxd_fastpath.c
nikhil-bhupale: Yes, for this release we are not destroying threads for decrement.

pxd_fastpath.c
sulakshm: ok context is always valid... it is never invalid, it is the number of cpu threads under that may change.

pxd_fastpath.c
```diff
@@ -780,7 +865,7 @@ void fastpath_queue_work(struct kthread_work* work, bool completion)
 		struct pxfpcontext_per_node *c = &pxfpctxt[node];
 		if (c->valid) {
 			cpuid = balanceIO(c, cpuid, completion);
-			worker = c->fpworker[cpuid & MAX_PXFP_WORKERS_PER_NODE_MASK];
+			worker = c->fpworker[cpuid % MAX_PXFP_WORKERS_PER_NODE];
```
sulakshm, line 868: mod is an expensive call relatively compared to bit ops. dont see a way out if not a power of 2. But maybe there is a kernel macro that improves this is MAX_FP_WORKERS_PER_NODE turns out to be a power of 2.

pxd_fastpath.c
sulakshm: ok publish next set of IMPROVEMENTs part of the PR/Design doc/notes etc so Mgmt and eng know about this.

PR #381, opened 2025-07-30
PWX-45726 Move pxd_timeout_msecs parameter to root level

nikhil-bhupale:
pxd_timeout_msecs parameter is added at device which requires device to be created to update this parameter. Move this to root level

[root@init500-d1 ~]# cat /sys/devices/pxd/
power/             pxd_num_fpthreads  timeout            uevent

### DIFF
--- a/pxd.c
+++ b/pxd.c
@@ -1941,10 +1941,47 @@ static ssize_t pxd_num_fpthreads_store(struct device *dev, struct device_attribu
 	return count;
 }
 
+static ssize_t pxd_timeout_show(struct device *dev,
+			 struct device_attribute *attr, char *buf)
+{
+	return sprintf(buf, "%u\n", pxd_timeout_secs);
+}
+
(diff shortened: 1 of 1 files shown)


pxd.c
```diff
@@ -1996,41 +2033,6 @@ static ssize_t pxd_minor_show(struct device *dev,
 			(unsigned long long)pxd_dev->minor);
 }
 
-static ssize_t pxd_timeout_show(struct device *dev,
-			 struct device_attribute *attr, char *buf)
-{
-	return sprintf(buf, "%u\n", pxd_timeout_secs);
-}
-
-static ssize_t pxd_timeout_store(struct device *dev, struct device_attribute *attr,
```
sulakshm, line 2005: dont remove any of these... this will break the control plane that will set timeout to zero during upgrades and some particular use cases to speed up removal of kernel module.

The PR should enable setting this under top level pxd device tree instead of through individual pxd block device minor path.

